PR #2, opened 2019-10-30
feature: Add EKS terraform

h314to:
There are still a few steps missing on `setup`, and I need to add the installation docs, but everything is working nicely. I'll open a PR with the remaining stuff after this is merged.


mrfyda:
can you change the default tag to codacy? (instead of beta)
Ideally we want to keep this right?

h314to:
> can you change the default tag to codacy? (instead of beta)
> Ideally we want to keep this right?

Yes, It should be just a matter of changing the default for `project_name` and `project_slug`. It's not hardcoded anywhere.

PR #5, opened 2019-10-31
feature: update helm dependencies on cluster and create changelog artifacts

joaocosta-codacy:


.circleci/getChangelogs.sh
```diff
@@ -0,0 +1,57 @@
...
+rm -rf ./changelogs
+mkdir changelogs
+for key in $DEPENDENCIES
+do
+    OLD_VERSION=$(yq r $OLD_LOCK_FILE dependencies -j | jq ".[] | select(.name==\"$key\").version" | sed "s/\"//g")
+    NEW_VERSION=$(yq r $NEW_LOCK_FILE dependencies -j | jq ".[] | select(.name==\"$key\").version" | sed "s/\"//g")
+    [ "$OLD_VERSION" != "$NEW_VERSION" ] && cloneRepository $key $OLD_VERSION $NEW_VERSION 
+done
+
+find ./changelogs -size 0 -delete # delete empty changelogs
```
joaocosta-codacy: the -empty option doesn't work on alipine linux. I think it must be the gnu tools that are installed are older than the ones we have on our machines.

.circleci/getChangelogs.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+REQUIREMENTS_FILE="codacy/requirements.yaml"
```
joaocosta-codacy: Following your suggestion @paulopontesm , helm dep up does not propagate the git info that exists on requirements.yaml on to requirements.lock. 
We must use the git info stored in the requirements.yaml after all.

PR #6, opened 2019-10-31
feature: start/destroy digital ocean cluster

joaocosta-codacy:


paulopontesm:
rename commits and PR to reflect what this is changing

h314to:
There are a few conflicts you need to tackle. Ping me after they are fixed.

I'm not a big fan of having the doks cluster code hidden away in a CI-specific dot dir. I guess it is ok for now, so you won't be blocked by this. I'll refactor it later.

PR #8, opened 2019-10-31
fix: Add docker creadentials

h314to:


h314to:
Fails on CI because it doesn't have a `.circleci/config.yaml`

PR #12, opened 2019-11-05
Feature/fluentd daemonset FT-8272

DReigada:


docs/configuration/logging.md
```diff
@@ -0,0 +1,10 @@
+## Logging
```
mrfyda: can you link this from the docs/README?

DReigada: There was already a link. That's why I chose this location for the file 😄

PR #22, opened 2019-11-14
Feature/halt workflow on failed deployments

joaocosta-codacy:
If a deployment is somehow broken after the helm install, the build should fail.
The reason could be the bump up of a chart version which might not get all pods up and running. Thus, we should catch this as early as possible. Otherwise, this sort of issues would go undetected until some tests fail for non obvious reasons.


.circleci/config.yml
```diff
@@ -39,6 +39,27 @@ jobs:
...
+    working_directory: ~/workdir/
+    steps:
+      - attach_workspace:
+          at: ~/
+      - run:
+          name: "Setup DO Credentials"
+          command: doctl auth init -t $DO_TOKEN &>/dev/null
+      - run:
+          name: "Validate deployment status"
+          command: bash ./.circleci/validateDeploymentStatus.sh
```
paulopontesm: you don't need the bash here.. you've put a shebang on the script

paulopontesm: I would also like it more to have this inline. But if you like it more this way its ok for me too

PR #74, opened 2019-12-09
fix: helm publications on workflows

joaocosta-codacy:
hourly pipeline no longer pushes to chartmuseum. 
nightly pipeline pushes only to nightly channel

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -129,7 +129,15 @@ references:
             helm dep up ./codacy
             echo "Adding 'https://charts.codacy.com/${CHANNEL}'"
             helm repo add --username "${CHARTS_REPO_USER}" --password "${CHARTS_REPO_PASS}" codacy https://charts.codacy.com/${CHANNEL}
-            helm push ./codacy codacy
+
+            if "$CHANNEL" == "stable"; then
+                helm push ./codacy codacy --version "$(cat .version)"
+            elif "$CHANNEL" == "nightly"; then
+                helm push ./codacy codacy --version "$(cat .version)-NIGHTLY"
+            else
+                helm push ./codacy codacy --version "$(cat .version)-SNAPSHOT"
+            fi
+            
       - persist_to_workspace:
           root: ~/
           paths:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -237,6 +240,12 @@ jobs:
       <<: *incubator_helm_channel
     <<: *helm_push
 
+  helm_push_nightly:
+    <<: *default_doks_image
+    environment:
+      <<: *nightly_helm_channel
+    <<: *helm_push
```
paulopontesm: Won't you hit the same problem here? The second night the version given by git-version is the same because de commit is the same.

`helm push --version "$(cat .version)-NIGHTLY"`

joaocosta-codacy: fixed in the next commit.

.circleci/config.yml
```diff
@@ -126,7 +129,15 @@ references:
             helm dep up ./codacy
             echo "Adding 'https://charts.codacy.com/${CHANNEL}'"
             helm repo add --username "${CHARTS_REPO_USER}" --password "${CHARTS_REPO_PASS}" codacy https://charts.codacy.com/${CHANNEL}
-            helm push ./codacy codacy
+
+            if "$CHANNEL" == "stable"; then
```
paulopontesm: this looks very hacky. This entire anchor just stopped being generic this way.
maybe just change the `.version` file in the workflow before calling this step?

joaocosta-codacy: I agree. I have extracted this into a separate job that we call in the workflow.

.circleci/config.yml
```diff
@@ -350,7 +375,7 @@ workflows:
           name: helm_lint
           cmd: helm lint codacy/
           requires:
-            - codacy/checkout_and_version
+            - odacy/checkout_and_version
```
paulopontesm: this seems wrong

.circleci/config.yml
```diff
@@ -126,7 +129,8 @@ references:
             helm dep up ./codacy
             echo "Adding 'https://charts.codacy.com/${CHANNEL}'"
             helm repo add --username "${CHARTS_REPO_USER}" --password "${CHARTS_REPO_PASS}" codacy https://charts.codacy.com/${CHANNEL}
-            helm push ./codacy codacy
+            helm push ./codacy codacy --version "$(cat .version)"
```
paulopontesm: This replaces the `version`  in the "./codacy/Chart.yaml" file. You are already doing this with the ytool above. I don't think this change is needed.

.circleci/config.yml
joaocosta-codacy: Typo. Fixed.

PR #75, opened 2019-12-09
fix: Setting codacy-api url after deployment via helm

joaocosta-codacy:


.circleci/config.yml
```diff
@@ -85,6 +85,10 @@ references:
       - deploy:
           name: Install Codacy
           command: make -C .doks/ deploy_to_doks
+      - deploy:
+          name: Update codacy-api ip address
+          command: |
+            ip=$(kubectl get services codacy-api --namespace codacy-nightly | awk '{print $4}' | tail -n +2); helm upgrade $RELEASE_NAME --set codacy-api.url=$ip
```
paulopontesm: I think its safer if you use `--jsonpath` here to get the ip for the service.

joaocosta-codacy: good call :-) i have replaced the command in the next commit.

.circleci/config.yml
```diff
@@ -85,6 +85,10 @@ references:
       - deploy:
           name: Install Codacy
           command: make -C .doks/ deploy_to_doks
+      - deploy:
+          name: Update codacy-api ip address
+          command: |
+            ip=$(kubectl get services codacy-api --namespace $NAMESPACE -o=jsonpath='{.status.loadBalancer.ingress[0].ip}'); helm upgrade $RELEASE_NAME --set codacy-api.url=$ip
```
paulopontesm: break line here

PR #76, opened 2019-12-09
Add resources requests

DReigada:


docs/README.md
```diff
@@ -34,7 +34,8 @@ helm delete codacy
 
 
 ## Requirements
-[All requirements](requirements/index.md)
+- [All requirements](requirements/index.md)
```
paulopontesm: Do you need to split these documents?

paulopontesm: I think we should start simple and putting everything in a single doc and split if necessary whenever is needed

DReigada: I felt that an index file would just be a listing of all the documents in its directory. Therefore the name index.

docs/requirements/resources.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Memory
+
+`workerResources.requests.memory + (pluginResources.requests.memory * 4)`
+
+
+### Number of concurrent analysis
+`workers.genericMax + workers.dedicatedMax`
+
+### Total resources
```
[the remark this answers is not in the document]
DReigada: That would probably be a good idea, but we are still no sure on how much resources we need. @paulopontesm  any input on this?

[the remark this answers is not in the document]
DReigada: Done

PR #78, opened 2019-12-10
feature: commit updated requirements.lock and re-deploy

joaocosta-codacy:


.circleci/config.yml
```diff
@@ -485,10 +503,62 @@ workflows:
           context: CodacyDO
           requires:
             - setup_qa_environment_nightly
-      - helm_push_nightly:
+      - commit_requirements:
```
joaocosta-codacy: The key part here is that commit_requirements does not persist to the workspace so the next checkout_and_version will be a fresh checkout.

joaocosta-codacy: To clarify, this will commit to the 'dev' branch

paulopontesm:
fix the commit messages please

PR #83, opened 2019-12-10
Fix/codacy url

joaocosta-codacy:


.circleci/config.yml
```diff
@@ -87,7 +87,7 @@ references:
       - <<: *doctl_authenticate
       - deploy:
           name: Install Codacy
-          command: make -C .doks/ deploy_to_doks
+          command: make -C .doks/ deploy_to_doks RELEASE_NAME=$RELEASE_NAME NAMESPACE=$NAMESPACE
```
paulopontesm: I don't think you need to pass the env variables

joaocosta-codacy: Fixed

PR #87, opened 2019-12-11
Feature: validation on multiple nightly clusters

joaocosta-codacy:
Currently only working for 1.14 and 1.15.

1.16 has a couple of issues with dependencies using apiversion beta1 or beta2 which we need to sort out first before we can include a 1.16 cluster in the workflow.


paulopontesm:
Can you add [WIP] tag plz?

.circleci/config.yml
```diff
@@ -76,7 +84,7 @@ references:
       - deploy:
           name: "Purge Cluster"
           command: |
-            make -C .doks/ remove_codacy  
+            set +e; make -C .doks/ remove_codacy; set -e
```
paulopontesm: why do you need the `set -e` at the end?

.circleci/config.yml
```diff
@@ -25,15 +24,24 @@ references:
...
     NAMESPACE: codacy-nightly
     NODE_TYPE: s-20vcpu-96gb
-    K8S_VERSION: 1.15
     NUM_NODES: 4
-  
+
+  nightly_environment_1_14: &nightly_environment_1_14
+    <<: *nightly_base_environment
+    DO_TF_WORKSPACE: nightly-1-14
+    DOKS_CLUSTER_NAME: codacy-doks-cluster-nightly-1-14
```
paulopontesm: We need to be creating/manage the changes of these clusters in this repo and pipeline also

.circleci/config.yml
```diff
@@ -187,6 +189,11 @@ references:
       - run:
           name: Set cluster
           command: doctl kubernetes cluster kubeconfig save "$DOKS_CLUSTER_NAME" --set-current-context
+      - run:
+          name: "Create namespace"
+          command: |
+            kubectl delete namespace $NAMESPACE
```
paulopontesm: Wont you delete everything if you do this?? Including all the data (pvcs) in the k8s dev installtion?

paulopontesm:
please fix the commit messages

.circleci/config.yml
joaocosta-codacy: They only get deleted if you --purge

.circleci/config.yml
joaocosta-codacy: How do you suggest we create them on the pipelines? 
Just verify if they do not exist, and if that's the case create them, otherwise continue?

joaocosta-codacy: Because otherwise we are not using the jobs as part of the pipelines

.circleci/config.yml
paulopontesm: purge is only a thing from helm https://v2.helm.sh/docs/helm/#helm-delete:
image

paulopontesm: kubectl dosen't have purge

paulopontesm: If you delete the namespace everything inside the namespace is deleted (pvcs included). I am not sure if you will get the same PVs once you create e everything again.

.circleci/config.yml
paulopontesm: `Just verify if they do not exist, and if that's the case create them, otherwise continue?` -> yes

paulopontesm: and change if anything is different... 
basically, make sure that the infrastructure matches the configuration in the repo

PR #108, opened 2019-12-16
feature: added jobs to start clusters

joaocosta-codacy:


.circleci/config.yml
```diff
@@ -186,6 +203,11 @@ jobs:
       <<: *dev_environment
     <<: *deploy_to_cluster
 
+  start_doks_nightly_cluster:
+    environment:
+      <<: *nightly_environment
```
paulopontesm: will this also start the environments for 1.14 and 1.15?

PR #118, opened 2019-12-17
Docs/improve docs v3

paulopontesm:


docs/installation/uninstall.md
```diff
@@ -0,0 +1,21 @@
...
+```
+
+Note that the deletion of `pvc`s in the above command has to run in the background
+due to a cyclic dependency in one of the components. If you are unsure of these
+commands' effect please run the each of the `bash` subcommands and validate their output, viz.
+
+```bash
+echo "PVCs to delete:"
+kubectl get pvc -n codacy -o jsonpath='{.items[*].metadata.name}'
+echo "PODs to delte:"
```
joaocosta-codacy: PODS to delete

paulopontesm: uau.. well spotted

PR #135, opened 2019-12-20
bump: Use 4.114.0-SNAPSHOT.efe990e5e for codacy-api

machadoit:


codacy/requirements.lock
```diff
@@ -13,39 +13,39 @@ dependencies:
   version: 0.3.0
 - name: portal
   repository: https://charts.codacy.com/stable
-  version: 4.29.3
+  version: ^4.28.0
```
joaocosta-codacy: The lock file should not have semver versions. It should resolve to absolute versions. Which helm version are you using?

PR #137, opened 2019-12-21
[WIP] docs: improve day to day logging documentation

paulopontesm:
still need to test this


paulopontesm:
We don't want to recommend elastic inside the cluster

PR #138, opened 2019-12-23
Merge Release 0.3.1

joaocosta-codacy:


codacy/requirements.lock
```diff
@@ -48,4 +48,4 @@ dependencies:
   repository: https://charts.codacy.com/stable
   version: 4.3.5
 digest: sha256:153c56c4e6398b2eecaa91f305189858583c09c9d23143386ecae78641dc1727
-generated: "2019-12-20T14:02:13.295509Z"
+generated: "2019-12-20T16:16:25.799537Z"
```
jonathancarvalheiro: Should this file be committed? 
Seems to be a generated file

.doks/Makefile
```diff
@@ -1,17 +1,36 @@
...
 	$(MAKE) -C ../ update_dependencies
-
 	helm upgrade --install ${RELEASE_NAME} ../codacy/ \
-	 	-f values.yaml \
+		--recreate-pods \
+		-f values.yaml \
+		-f providers.yaml \
+		--namespace ${NAMESPACE} \
+		--set global.imagePullSecrets[0].name=docker-credentials \
+		--set global.play.cryptoSecret=$${SHARED_PLAY_CRYPTO_SECRET} \
```
jonathancarvalheiro: For curiosity what is the difference between the $ and $$?

joaocosta-codacy: Single dollar sign -> makefile var
Double dollar sign -> shell (env) var
:-)

codacy/requirements.lock
joaocosta-codacy: This file should be checked in because this is what helm will use to determine the versions of all the components to be installed. As part of a release, we do not perform a helm dep up command so we want to use the "frozen" versions, i.e. the versions in the .lock file.

jonathancarvalheiro: understood, thanks :D

PR #140, opened 2019-12-23
fix: Fix README images

DReigada:
Codacy <3 k8s


DReigada:
git-version doesn't like my branch name :(

PR #150, opened 2019-12-30
feature: compatibility with removal of resultsdb FT-6848

pedrocodacy:
You can review this as-is, but I'll still do a commit on top bumping the component versions for the ones that use results db (with the changes that use the changed values)


pedrocodacy:
Bumped the versions of all components that use resultsDB. 
These are not published as stable at the time of writing this comment, so please wait for my go-ahead before you merge this PR

PR #157, opened 2020-01-03
fix: evolutions for results2017 FT-6848

pedrocodacy:
At the time of writing this comment this version of `ragnaros` is not yet published. 
Please wait for a comment saying that this is good to merge before doing so


pedrocodacy:
`ragnaros-15.1.1` is published, we can now merge this

PR #168, opened 2020-01-08
fix: Freeze worker version.

h314to:


codacy/values.yaml
```diff
@@ -273,6 +273,10 @@ worker-manager:
       tolerations: []
       nodeSelector: {}
 
+    codacy:
+      worker:
+        image: 5.56.0
```
paulopontesm: this needs to be updated automatically in the pipeline when there is a release

joaocosta-codacy: do you mean that this would be another helm --set ?
does circleci support explicitly asking for parameter values? so that when you cut the release, you can input the value of the worker image that you want to use? I can not find that anywhere :/

paulopontesm: no.. with the ytool. This needs to have the same value as the engine.

joaocosta-codacy: got it.

PR #173, opened 2020-01-08
fix: remove results references and bump wm

pedrocodacy:
Please wait for worker-manager 8.11.2 to be published. I'll comment in this PR when it is ready


pedrocodacy:
WM 8.11.2 is published

pedrocodacy:
did a new commit on top bumping ragnaros to the latest version

PR #181, opened 2020-01-10
fix: use semantic versions for all components

joaocosta-codacy:
We want to go back to using regexes for versions.
The reason is that our hourly QA tests and hourly and nightly rollout validations will need to deploy the latest versions of each component to know whether there have been any breaking changes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -384,12 +384,15 @@ jobs:
           command: |
             git config --global user.email "circleci@codacy.com"
             git config --global user.name "codacycircleci"
+            ssh-keyscan github.com >> ~/.ssh/known_hosts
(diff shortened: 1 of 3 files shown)


joaocosta-codacy:
In fact, as per https://devhints.io/semver
Do we want to automatically bump to the highest minor version, or to the highest patch version?
If we want highest patch versions, then i need to change the syntax on all the versions to use a ~ instead of a ^.

bmbferreira:
Correct me if I'm wrong, but by updating minor versions automatically, if I'm on a call with a customer and I know that they have the version X of the chart I have to check the components individually to check the minor version that they are currently running. :/

joaocosta-codacy:
Those expressions are only used when cutting the release. Meaning, if you get the chart (version 1.2.3), afaik you will get the components with the versions at the time that the chart release was made.

PR #184, opened 2020-01-14
added guide for release process for self hosted enterprise

joaocosta-codacy:


docs/release/release.md
```diff
@@ -0,0 +1,34 @@
+## Preparing the release
+
+1. Create a release branch in the form of `release/[version]`
```
paulopontesm: instead on [version] i think DD-MM-YYYY-HH:MM or something would be better. The version is generated with git-version. This might trick you

docs/release/release.md
```diff
@@ -0,0 +1,34 @@
...
+2. Freeze the versions of each component on the requirements.yaml file to this branch.
+3. Run `helm dep up`.
+4. Commit both `requirements.yaml` and `requirements.lock` to the branch. The commit message must be "`release: start release [version]`".
+5. Create a pull request to `master`.
+6. Wait for the merge. :)
+7. Tag the merge with `RELEASE-[version]`
+8. This tag will automatically trigger a release candidate build. You chart will be deployed to `release.dev.codacy.org` (`codacy-doks-cluster-release` cluster in digital ocean).
```
paulopontesm: @h314to can you confirm this part of the tags?

docs/release/release.md
```diff
@@ -0,0 +1,34 @@
+## Preparing the release
+
+1. Create a release branch in the form of `release/[version]`
+2. Freeze the versions of each component on the requirements.yaml file to this branch.
```
paulopontesm: Can you add some instructions on how to get the latest versions here?
Like we do on circleci

docs/release/release.md
joaocosta-codacy: Done

docs/release/release.md
joaocosta-codacy: Done

docs/release/release.md
```diff
@@ -0,0 +1,72 @@
+## Preparing the release
```
paulopontesm: This should be the title (only one `#` )

docs/release/release.md
```diff
@@ -0,0 +1,72 @@
...
+
+1. Create a release branch in the form of `release/[DD-MM-YYYY]`
+2. Freeze the versions of each component on the requirements.yaml file to this branch. Please see the [section below](##getting-the-latest-versions-of-all-components).
```
paulopontesm: I think you can put the instructions in line here, instead of a different section

PR #216, opened 2020-01-27
feature: Add Akka session secret

andreaTP:
I don't know were `SHARED_PLAY_CRYPTO_SECRET` is coming from and not sure where to add `{SHARED_AKKA_SESSION_SECRET`


docs/install.md
```diff
@@ -26,6 +26,8 @@ global:
     backendUrl: "http://codacy.mycompany.com" # This value is important for VCS configuration and badges to work
   play:
     cryptoSecret: "PLEASE_CHANGE_ME" # Generate one with `openssl rand -base64 32 | tr -dc 'a-zA-Z0-9'`
+  akka:
+    sessionSecret: "PLEASE_CHANGE_ME" # Generate one with `openssl rand -base64 32 | tr -dc 'a-zA-Z0-9'`
```
andreaTP: why we are indicating a different why of obtaining the secret other than: `cat /dev/urandom`
can we try to be consistent?

joaocosta-codacy:
SHARED_PLAY_CRYPTO_SECRET comes from the CodacyDO context in circleci.
I'd say that SHARED_AKKA_SESSION_SECRET should be put in the context as well.

.doks/Makefile
```diff
@@ -13,6 +13,7 @@ deploy_to_doks: set_cluster_context
 		--namespace ${NAMESPACE} \
 		--set global.imagePullSecrets[0].name=docker-credentials \
 		--set global.play.cryptoSecret=$${SHARED_PLAY_CRYPTO_SECRET} \
+		--set global.akka.sessionSecret=$${SHARED_AKKA_SESSION_SECRET} \
```
paulopontesm:
```suggestion
		--set global.akka.sessionSecret=$${SHARED_PLAY_CRYPTO_SECRET} \
```

Since this is a dev environment I think we can use the same "secret" fall everything

.doks/Makefile
```diff
@@ -33,6 +34,7 @@ deploy_to_doks_from_chartmuseum: set_cluster_context
...
```
paulopontesm:
```suggestion
		--set global.akka.sessionSecret=$${SHARED_PLAY_CRYPTO_SECRET} \
```

Since this is a dev environment I think we can use the same "secret" fall everything

docs/install.md
paulopontesm: You are right.. can you change all these `openssl` ?

.doks/Makefile
andreaTP: I added the env to the circle ci pipeline

docs/install.md
andreaTP: done

.doks/Makefile
andreaTP: akka session secret has a minimum required length that is really different from the play one, I would prefer to keep them separated.

paulopontesm:
Can you just set on the “requirements.yaml” the mininimum version where these credentials are needed and fix the conflicts?

andreaTP:
Opened #233 on top of latest master

PR #218, opened 2020-01-28
fix: unfreezing versions on master using semver

joaocosta-codacy:


.doks/Makefile
```diff
@@ -8,7 +8,6 @@ DEPLOYMENTS=$(shell kubectl get deployments -n "${NAMESPACE}" | awk '{print $$1}
 deploy_to_doks: set_cluster_context
 	$(MAKE) -C ../ update_dependencies
 	helm upgrade --install ${RELEASE_NAME} ../codacy/ \
-		--recreate-pods \
```
joaocosta-codacy: This still raises problems whenever the components bump their versions up, leaving the deployment in a messed up state. Hence, rotate_pods was re-added to the release workflow (see config.yml)

paulopontesm: this needs to work. What is the problem?

joaocosta-codacy: For some reason, when using recreate-pods, components which had their version bumped up will be stuck. The deployment will be left in a weird transitive state with n pods of the old version and n pods of the new version, some running, some pending. :-( 
This is the reason why we had created the 'rotate_pods' job which circumvents the issue.

joaocosta-codacy: reverted back to having recreate-pods.

joaocosta-codacy:
removed requirements-dev.yaml

PR #221, opened 2020-01-29
fix: changelog git tag command fixed

joaocosta-codacy:


.circleci/getChangelogs.sh
```diff
@@ -55,14 +55,5 @@ do
     fi
 done
 
-count=$(find ./changelogs -maxdepth 1 -type f | wc -l | awk '{print $1}')
```
joaocosta-codacy: We only export 1 changelog file so this is not needed anymore.

PR #225, opened 2020-01-31
feature: Add new configuration value for commit diff bucket

afsmeira:


afsmeira:
The versions are not yet stable (were only published as snapshots, to test in DEV). I'll add the correct version now, but I'll only run that command and finish the PR when there are stable versions.

codacy/values.yaml
```diff
@@ -70,6 +70,8 @@ global:
         name: "coverage"
       cliReports:
         name: "cli-reports"
+      commitDiff:
+        name: "commit-diff"
```
rtfpessoa: Shouldn't we use `https://github.com/helm/charts/tree/master/stable/minio#configuration` to create the buckets instead of having special code to do it?

DReigada: For compatibility reasons with legacy-enterprise we still need to create the buckets :/

rtfpessoa: not really, if it is running it should exist by now.

paulopontesm: What do you mean @rtfpessoa ? 
In legacy-enterprise we are creating the buckets with the component, right?

rtfpessoa: yup, but we could do this same improvement there. This is just a variable, isn't this supported by the docker itself?

Still even if it is not doing it now should make it easier for the future.

paulopontesm: ok.. so the suggestion is to create all the buckets when minio is created, by putting the ".values.minio.buckets = xxxx" in the values file

image

If the bucket is created, the component is going to detect that and just use that bucket instead of creating one.

Is that it?

rtfpessoa: yes

afsmeira: @rtfpessoa, @joaocosta-codacy is implementing your request in another PR, on top of this one. I'll resolve this comment.

PR #243, opened 2020-02-06
fix: added minio buckets

joaocosta-codacy:


codacy/values.yaml
```diff
@@ -369,6 +369,21 @@ minio:
     size: 20Gi
   minioConfig:
     region: "eu-west-1"
+  buckets:
+  - name: "analysis"
+    policy: "public"
```
joaocosta-codacy: as per the chart's documentation, the available policies are none|download|upload|public.

PR #249, opened 2020-02-10
fix-hotspots-typo

hjrocha:


dan-codacy:
@hjrocha There's an issue with CircleCi and PRs coming from forks, so your helm_lint check will never complete. I had to branch instead of forking in order to get the check to run

pauloribeiro-codacy:
Already fixed in master.

PR #257, opened 2020-02-12
feature: rotate pods to avoid nfs provisioning issues

joaocosta-codacy:


.circleci/config.yml
```diff
@@ -125,6 +124,14 @@ references:
           name: Install Codacy
           command: make -C .doks/ deploy_to_doks_from_chartmuseum RELEASE=$RELEASE NAMESPACE=$NAMESPACE VERSION=$(cat .version) HELM_REPOSITORY=$HELM_REPOSITORY
 
+  rotate_pods: &rotate_pods
```
paulopontesm: can you put this right after the helm upgrade? on the makefile?
Its stays just in one place and its easier to delete later

joaocosta-codacy: sure! :) that's actually much simpler!

PR #262, opened 2020-02-17
refactor: expose worker version patching on makefile

joaocosta-codacy:
We no longer need to patch this worker version if we expose it through the makefile.
The default version is the stable tag.

### DIFF
--- a/.doks/Makefile
+++ b/.doks/Makefile
@@ -3,6 +3,7 @@ RELEASE_NAME?="codacy"
 NAMESPACE?="codacy"
 HELM_REPOSITORY?=codacy-incubator
 DEPLOYMENTS?=$(shell kubectl get deployments -n "${NAMESPACE}" | awk '{print $$1}' | tail -n +2 | grep -v minio)
+WORKER_VERSION?=$(shell grep "engine" -A 2 codacy/requirements.lock | grep version | cut -d : -f 2 | tr -d '[:blank:]')
 
 define call_helm_install
 	helm upgrade --install ${1} ${3} \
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -1,16 +1,10 @@
-ENGINE_VERSION:=$(shell grep "engine" -A 2 codacy/requirements.lock | grep version | cut -d : -f 2 | tr -d '[:blank:]')
 .PHONY: setup_helm_repos
 setup_helm_repos:
 	helm repo add codacy-stable https://charts.codacy.com/stable
 	helm repo add codacy-unstable https://charts.codacy.com/unstable
 	helm repo add codacy-incubator https://charts.codacy.com/incubator
 	helm repo add codacy-external https://charts.codacy.com/external
 
-.PHONY: update_worker_version
```
paulopontesm: don't you need this for the release process?

joaocosta-codacy: Not anymore. All the workflows in this chart install the stable components so we no longer need to update this. The default worker image is stable as well. :)

joaocosta-codacy: This is now done on the doks makefile itself (See above)

.doks/Makefile
```diff
@@ -3,6 +3,7 @@ RELEASE_NAME?="codacy"
 NAMESPACE?="codacy"
 HELM_REPOSITORY?=codacy-incubator
 DEPLOYMENTS?=$(shell kubectl get deployments -n "${NAMESPACE}" | awk '{print $$1}' | tail -n +2 | grep -v minio)
+WORKER_VERSION?=$(shell grep "engine" -A 2 codacy/requirements.lock | grep version | cut -d : -f 2 | tr -d '[:blank:]')
```
joaocosta-codacy: this way we always use the engine version as the worker version by default. no longer using rolling tags.

Makefile
paulopontesm: still desent work for the release process... we need to release with this hardcoded

joaocosta-codacy: Got it. 👍

PR #266, opened 2020-02-18
[wip] bump: helm version

joaocosta-codacy:


.doks/doks-cluster/config.tf
```diff
@@ -23,8 +23,8 @@ provider "helm" {
   # tiller is only installed if you actually do something with this provider, such as installing a chart
   install_tiller  = "true"
   service_account = "tiller"
-  tiller_image    = "gcr.io/kubernetes-helm/tiller:v2.15.0"
-  
+  tiller_image    = "gcr.io/kubernetes-helm/tiller:v2.16.3"
```
paulopontesm: This should be done with `helm init --upgrade`. This way it stays dependent on the version of the client in the ci-do docker image

PR #267, opened 2020-02-19
Fix/make dev explicit

paulopontesm:


.doks/Makefile
```diff
@@ -26,14 +26,10 @@ define call_helm_install
 endef
 
 .PHONY: deploy_to_doks
-deploy_to_doks: set_cluster_context
+deploy_to_doks_from_local_dir: set_cluster_context
 	$(MAKE) -C ../ update_dependencies
 	$(call call_helm_install,${RELEASE_NAME},${NAMESPACE},../codacy/,${CODACY_URL})
 
-.PHONY: deploy_to_doks_atomic
```
joaocosta-codacy: We need this on the codacy-website job.
The purpose was to have an helm upgrade with no other operations before it, so that as soon as the polling completes we immediately start installing.

PR #268, opened 2020-02-20
feature: ensure clusters are started on workflows

joaocosta-codacy:
With this we can ensure that the clusters are up and running with tiller (specific version) installed

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -100,6 +100,17 @@ references:
     attach_workspace:
           at: ~/
 
+  start_cluster: &start_cluster
+    steps:
+      - <<: *attach_workspace
+      - <<: *doctl_authenticate
+      - run:
+          name: "Start Cluster"
+          command: make -C ./.doks/doks-cluster cluster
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -100,6 +100,17 @@ references:
     attach_workspace:
           at: ~/
 
+  start_cluster: &start_cluster
```
paulopontesm:
```suggestion
  create_cluster: &create_cluster
```

.circleci/config.yml
```diff
@@ -100,6 +100,17 @@ references:
...
+    steps:
+      - <<: *attach_workspace
+      - <<: *doctl_authenticate
+      - run:
+          name: "Start Cluster"
+          command: make -C ./.doks/doks-cluster cluster
+      - run:
+          name: "Populate Cluster"
+          command: make -C ./.doks/doks-cluster populate_cluster
```
paulopontesm: This step and the one before this can be merged in the makefile

PR #275, opened 2020-02-21
refactor: decrease number of component replicas

joaocosta-codacy:


joaocosta-codacy:
This applies only to our digital ocean installs

PR #276, opened 2020-02-21
bump: ci-do image version

joaocosta-codacy:


joaocosta-codacy:
This will move us forward into helm 2.16.3 and terraform 0.12.21

PR #284, opened 2020-02-24
[wip] Terraform cluster as part of workflows

joaocosta-codacy:


.doks/doks-cluster/Makefile
```diff
@@ -4,13 +4,15 @@ TF_EXTRA_OPTS ?= TF_IN_AUTOMATION=1 AWS_SDK_LOAD_CONFIG=1 TF_CLI_CONFIG_FILE=$(T
 TF_PARALLELISM ?= 30
 TF ?= $(TF_EXTRA_OPTS) terraform
 WORKSPACE ?= ${DO_TF_WORKSPACE}
+NAMESPACE?=codacy
 VARS_FILE ?= deploy.tfvars
 
 # variables
-K8S_VERSION ?= 1.15
+K8S_VERSION ?= 1.14
```
paulopontesm: are we going back to 1.14?

joaocosta-codacy: this is the k8s version that was originally on the cluster and we were actually passing this value onto the makefile.
:)
if we want to move to to 1.15 then we need to create a new cluster :-(

.doks/doks-cluster/Makefile
```diff
@@ -29,25 +31,24 @@ setup_vars:
...
 .PHONY: start_cluster
-start_cluster: set_cluster_context helm_init_upgrade
+start_cluster: set_cluster_context cluster helm_init_upgrade
 
 .PHONY: cluster
 cluster: setup_vars create_terraformrc
 	echo '1' | $(TF) init -input=false -reconfigure
 	$(TF) workspace select $(WORKSPACE)
 	$(TF) apply -parallelism=$(TF_PARALLELISM) -var-file=$(VARS_FILE) -input=false --auto-approve
-	$(shell doctl kubernetes cluster kubeconfig save $(DOKS_CLUSTER_NAME) --set-current-context)
```
joaocosta-codacy: We no longer provision helm through terraform so we no longer need this step nor the following terraform apply

.doks/doks-cluster/cluster.tf
```diff
@@ -5,43 +5,23 @@ resource "digitalocean_kubernetes_cluster" "codacy_k8s" {
 
   node_pool {
     name       = "codacy-doks-pool-${terraform.workspace}"
-    size       = var.node_type
-    node_count = var.num_nodes
+    size       = "s-2vcpu-2gb"
+    auto_scale = true
+    node_count = 1
```
joaocosta-codacy: The digital ocean kubernetes cluster terraform resource isn't great. It kind of enforces you to always have a default node_pool defined inside the cluster. The problem is that once that node pool is created, any change to it will cause terraform to want to destroy the cluster and provision a new cluster, instead of just a new node pool. 
Others have found the same problem [here](https://github.com/terraform-providers/terraform-provider-digitalocean/issues/350) and [here](https://github.com/terraform-providers/terraform-provider-digitalocean/issues/303).
The solution was to get the node pool size down to zero, then manually delete it through digital ocean's UI. This enforces the default node pool never to exist. Then we created another extra node pool (see below).

joaocosta-codacy: The digital ocean kubernetes cluster terraform resource isn't great. It kind of enforces you to always have a default node_pool defined inside the cluster. The problem is that once that node pool is created, any change to it will cause terraform to want to destroy the cluster and provision a new cluster, instead of just a new node pool. 
Others have found the same problem [here](https://github.com/terraform-providers/terraform-provider-digitalocean/issues/350) and [here](https://github.com/terraform-providers/terraform-provider-digitalocean/issues/303).
The solution was to get the node pool size down to zero, then manually delete it through digital ocean's UI. This enforces the default node pool never to exist. Then we created another extra node pool (see below).

.doks/doks-cluster/Makefile
```diff
@@ -29,25 +30,23 @@ setup_vars:
 	@echo "docker_password = [REDACTED]"
 	@echo "docker_password = \"$(DOCKER_PASSWORD)\"" >> $(VARS_FILE)
 
-	echo "k8s_version = \"$(DO_K8S_SLUG)\"" >> $(VARS_FILE)
-	echo "node_type = \"$(NODE_TYPE)\"" >> $(VARS_FILE)
-	echo "num_nodes = $(NUM_NODES)" >> $(VARS_FILE)
```
joaocosta-codacy: We no longer need this since we use an auto scalable node pool

.doks/doks-cluster/cluster.tf
```diff
@@ -5,43 +5,23 @@ resource "digitalocean_kubernetes_cluster" "codacy_k8s" {
...
-resource "kubernetes_secret" "do_token" {
-  count = var.cluster_only ? 0 : 1
-  metadata {
-    name = "digitalocean"
-    namespace = "kube-system"
-  }
-  data = {
-    "access-token" = var.digital_ocean_token
-  }
+resource "digitalocean_kubernetes_node_pool" "auto-scale-pool-01" {
```
joaocosta-codacy: This is the actual node pool that we use.
Due to shortcomings on the kubernetes terraform resource from digital ocean, if we want to change the node_type on this  pool, we will have to do so in multiple steps.
Changing the node_type for the environments on the config.yaml file will cause this node pool to be recreated (i.e. destroyed and created), effectively crushing to bits anything inside it. :'( 
What we can do is add a new node pool with the desired specification, then scale down this node pool. After all components have been moved to the new nodes, we can then scale up this node pool with the desired spec, and finally get rid of the second temporary node pool.
These were my findings, although this should not happen very often. I just want to raise my concern about this point, and that I am actually not happy with that approach, especially given that our terraform apply command is --auto-approve.
I am open to suggestions on how to actually prevent accidents from happening when changing the node_type.

.doks/doks-cluster/Makefile
```diff
@@ -29,25 +30,23 @@ setup_vars:
...
 
 .PHONY: start_cluster
-start_cluster: set_cluster_context helm_init_upgrade
+start_cluster: set_cluster_context cluster helm_init_upgrade
 
 .PHONY: cluster
 cluster: setup_vars create_terraformrc
 	echo '1' | $(TF) init -input=false -reconfigure
 	$(TF) workspace select $(WORKSPACE)
 	$(TF) apply -parallelism=$(TF_PARALLELISM) -var-file=$(VARS_FILE) -input=false --auto-approve
```
joaocosta-codacy: This terraform apply can be very dangerous if the NODE_TYPE of any environment on .circleci/config.yaml changes. See the last comment on .doks/doks-cluster/cluster.tf.

.doks/doks-cluster/Makefile
joaocosta-codacy: Also, by specifying this version here, with the terraform apply --auto-approve that we have below, we will guarantee that the cluster is up to date with regards to patch versions of k8s 1.14 automatically

joaocosta-codacy:
Superseeded by https://github.com/codacy/chart/pull/351.
I am dropping this PR.

PR #285, opened 2020-02-28
added microk8s installation docs

joaocosta-codacy:


docs/infrastructure/microk8s/install.md
```diff
@@ -0,0 +1,274 @@
...
+## Cookbook
+
+Assuming that you are starting from a blank slate, the first step is to install microk8s. Otherwise, please jump on to the [Configuring microk8s](###Configuring-microk8s) section.
+
+### Installing microk8s
+
+1. Ensure that you are working off the root folder: `cd /root`.
+2. Make sure the machine is up to date and also has the `nfs-common` package installed: `apt update && apt install nfs-common -y`.
+3. We must run configure the instance's list of hosts with localhost. This in turn will allow the correct resolution of the local microk8s cluster on the kubeconfig. We can do so by running `sed -i -e "1s/127.0.0.1 localhost.*/127.0.0.1 localhost $(hostname)/" /etc/hosts`.
+4. Traffic forwarding must be configured on the machine so that the local microk8s cluster pods can reach the internet. This can be done by running `iptables -P FORWARD ACCEPT` (more info on this subject can be [found here](https://github.com/ubuntu/microk8s/tree/1.12#my-pods-cant-reach-the-internet-but-my-microk8s-host-machine-can)).
```
andreaTP: This does not appear anymore in `master` of the project, can we instead indicate a minimum version supported of microk8s? (looks like the workaround is not needed > 1.14)

docs/infrastructure/microk8s/install.md
```diff
@@ -0,0 +1,274 @@
...
+5. Choose the microk8s release channel you want to install from.
+   1. The complete list of channels can be [found here](https://microk8s.io/docs/setting-snap-channel).
+   2. Assuming we want to install from the `stable` release channel, we export the following environment variable: `export MICROK8S_CHANNEL=stable`.
+6. We are finally in a state where we can install microk8s. That can be done via the `snap` package manager with: `snap install microk8s --classic --channel ${MICROK8S_CHANNEL}
+export PATH=$PATH:/snap/bin
+echo "--allow-privileged=true" >> /var/snap/microk8s/current/args/kube-apiserver
```
andreaTP: why is this change needed?

docs/infrastructure/microk8s/install.md
```diff
@@ -0,0 +1,274 @@
...
+
+* a machine running Ubuntu 18.04 LTS version.
+  * Further versions might work but we recommend using an LTS release.
+* a remote session (ssh) onto your machine of choice.
+
+## Cookbook
+
+Assuming that you are starting from a blank slate, the first step is to install microk8s. Otherwise, please jump on to the [Configuring microk8s](###Configuring-microk8s) section.
+
+### Installing microk8s
```
andreaTP: I would recommend to link to the upstream documentation for installing microk8s instead of rolling out our own.
Ideally we should solely maintain the diff that is needed for Codacy.(e.g. the installation of `nfs-common`).
https://microk8s.io/docs/
WDYT?

docs/infrastructure/microk8s/install.md
```diff
@@ -0,0 +1,274 @@
...
+   2. Assuming we want to install from the `stable` release channel, we export the following environment variable: `export MICROK8S_CHANNEL=stable`.
+6. We are finally in a state where we can install microk8s. That can be done via the `snap` package manager with: `snap install microk8s --classic --channel ${MICROK8S_CHANNEL}
+export PATH=$PATH:/snap/bin
+echo "--allow-privileged=true" >> /var/snap/microk8s/current/args/kube-apiserver
+microk8s.start && microk8s.status --wait-ready`
+
+### Configuring microk8s
+
+1. First, we must enable the following plugins on microk8s:
+   1. `sudo microk8s.enable dns && sudo microk8s.status --wait-ready --timeout 180`
```
andreaTP: I think that we do not need timeouts in manual steps.

docs/infrastructure/microk8s/install.md
```diff
@@ -0,0 +1,274 @@
...
+   2. `sudo microk8s.enable storage && sudo microk8s.status --wait-ready --timeout 180`
+   3. `sudo microk8s.enable ingress &&  sudo microk8s.status --wait-ready --timeout 180`
+   4. This will enable dns, storage and ingress on your microk8s cluster.
```
andreaTP: this is the title not a further point, I guess.

docs/infrastructure/microk8s/install.md
```diff
@@ -0,0 +1,274 @@
...
+2. Now that these plugins have been enabled, we should restart microk8s to ensure a proper bootstrap of the cluster: `microk8s.stop && microk8s.start && microk8s.status --wait-ready
+sleep 10`. Further information on this subject can be [found here](https://github.com/ubuntu/microk8s/issues/493#issuecomment-498167435).
```
andreaTP: we do not need to sleep here right?

docs/infrastructure/microk8s/install.md
```diff
@@ -0,0 +1,274 @@
...
+3. The plugins are now enabled and the cluster bootstrapped. However, we must still wait for some microk8s internals (dns, http, and ingress) plugins to be ready. Failing to do so can result in pods entering a `CrashLoopBackoff` state:
+   1. `microk8s.kubectl wait -n kube-system --for=condition=Ready pod -l k8s-app=kube-dns --timeout=300s`
```
andreaTP: I think we can squeeze all these commands in one that wait for everything and return to the user.
In a manual installation process we do not need to have timeout ideally and just let the user wait.

docs/infrastructure/microk8s/install.md
```diff
@@ -0,0 +1,274 @@
...
+   2. `microk8s.kubectl wait -n kube-system --for=condition=Ready pod -l k8s-app=hostpath-provisioner --timeout=300s`
+   3. `microk8s.kubectl wait -n default --for=condition=Ready pod -l app=default-http-backend --timeout=300s`
+   4. `microk8s.kubectl wait -n default --for=condition=Ready pod -l name=nginx-ingress-microk8s --timeout=300s`
+4. After runing these commands, we have ensured that dns, http, and nginx ingress are enabled and working properly inside the cluster.
```
andreaTP: this is a conclusion I guess it doesn't need the trailing `4.`

docs/infrastructure/microk8s/install.md
```diff
@@ -0,0 +1,274 @@
...
+
+### Installing Codacy
+
+#### Helm and Tiller
+
+1. First we must install `helm` onto the cluster. With `helm`, we can easily install charts and manage the lifecycle of the installed artifacts. One example is the ability to rollback a failed install.
```
andreaTP: This is an introduction, we don't need to have `1.`

docs/infrastructure/microk8s/install.md
```diff
@@ -0,0 +1,274 @@
...
+2. Assuming you would want to install `helm` version 2.16.3, export the following environment variable: `export HELM_VERSION=2.16.3`
```
andreaTP: Should we state that we are supporting Helm 2 and not 3?

docs/infrastructure/microk8s/install.md
```diff
@@ -0,0 +1,274 @@
...
+3. Run this script that will retrieve the desired `helm` version and install it onto the machine's `/usr/local/bin`: `HELM_PKG=helm-${HELM_VERSION}-linux-amd64.tar.gz
+wget https://get.helm.sh/$HELM_PKG
+tar xvzf $HELM_PKG
+mv linux-amd64/tiller linux-amd64/helm /usr/local/bin
```
andreaTP: this step require `su` privileges

docs/infrastructure/microk8s/install.md
```diff
@@ -0,0 +1,274 @@
...
+rm -rvf $HELM_PKG linux-amd64/`.
+   1. Two executables will get installed: `helm` and `tiller`. The former is responsible for resolving the configuration of the chart to be installed, while issuing the correct install commands onto the cluster. The latter is responsible for receiving the install commands issued by `helm`, as well as managing the lifecycle of the components that have been installed. `helm` is the client facing side, while `tiller` is the server/cluster facing side.
```
andreaTP: This is a very good explanation, I would have it before the installation phase

docs/infrastructure/microk8s/install.md
```diff
@@ -0,0 +1,274 @@
...
+4. We must now deploy `tiller` onto the cluster.
+   1. First we must set a `serviceaccount` and a `clusterrolebinding` for `tiller` on the microk8s cluster: `microk8s.kubectl create serviceaccount --namespace kube-system tiller
+microk8s.kubectl create clusterrolebinding tiller-cluster-rule --clusterrole=cluster-admin --serviceaccount=kube-system:tiller`
+   2. Then we can install `tiller`, which will be installed with the same `HELM_VERSION` that you have specified earlier: `helm init --service-account tiller`
+   3. We wait for tiller to be ready: `microk8s.kubectl -n kube-system wait --for=condition=Ready pod -l name=tiller --timeout=300s`
```
andreaTP: I might be wrong, but I think that for installing tiller we do not need dns etc.
We can probably install helm and tiller before enabling the "extra-services" in microk8s and than wait one time for all, WDYT?

docs/infrastructure/microk8s/install.md
```diff
@@ -0,0 +1,274 @@
...
+
+### Codacy chart
+
+1. Create a namespace inside the cluster onto which codacy will be deployed. `microk8s.kubectl create namespace codacy`
```
andreaTP: Should we let the user choose the namespace name or we have something that suggest us to use `codacy` ?

docs/infrastructure/microk8s/install.md
```diff
@@ -0,0 +1,274 @@
...
+2. The microk8s cluster must know some docker credentials so that the container docker images can be pulled onto the pods. These credentials must belong to the namespace where codacy is installed. You can do so by running: `microk8s.kubectl create secret docker-registry docker-credentials --docker-username=${DOCKER_USERNAME} --docker-password=${DOCKER_PASSWORD} --namespace codacy`. These credentials will be encoded inside the cluster.
+3. For codacy to work, we must set a file that which contains information like how many replicas of a given pod will codacy have, what are the cpu and memory limits for these pods, which secret tokens to use for the filestore, among others.
+   1. Define the following variables
+      1. `CODACY_PROTOCOL` - http or https protocol for your codacy installation.
+      2. `CODACY_URL` - desired URL for your codacy installation.
+      3. `CRYPTO_SECRET` - A secret...
+   2. Get your `microk8s-values.yaml` file ready by running `cat << END_VALUES > microk8s-values.yaml`
+   3. Paste the contents of [the microk8s-values.yaml file](####microk8s-values.yaml) written below.
+   4. Type `END_VALUES` to finish the inline `cat` onto the file.
```
andreaTP: I would let the user decide how to write the file and just provide an example and the needed filename here, WDYT?

docs/infrastructure/microk8s/install.md
```diff
@@ -0,0 +1,274 @@
...
+4. `helm` must know where it can get the charts from. You must add Codacy's chart museum to `helm`'s internal list: `
+helm repo add codacy-repo https://charts.codacy.com/stable/`
+5. Finally, you can install Codacy's chart by issuing the following `helm` command: `helm upgrade --install codacy codacy-repo/codacy --namespace codacy \
+  --recreate-pods --timeout 1200 --values microk8s-values.yaml`
```
andreaTP: I guess we don't need the timeout here

docs/infrastructure/microk8s/install.md
```diff
@@ -0,0 +1,274 @@
...
+
+```yaml
+global:
+  imagePullSecrets:
+    - name: docker-credentials
+  codacy:
+    url: "${CODACY_PROTOCOL}://${CODACY_URL}"
+    backendUrl: "${CODACY_PROTOCOL}://${CODACY_URL}"
+  akka:
+    sessionSecret: "${CRYPTO_SECRET}"
```
andreaTP: I think we should encourage users to use different secrets instead of reusing one.

docs/infrastructure/microk8s/install.md
```diff
@@ -0,0 +1,274 @@
...
+
+### Codacy chart
+
+1. Create a namespace inside the cluster onto which codacy will be deployed. `microk8s.kubectl create namespace codacy`
+2. The microk8s cluster must know some docker credentials so that the container docker images can be pulled onto the pods. These credentials must belong to the namespace where codacy is installed. You can do so by running: `microk8s.kubectl create secret docker-registry docker-credentials --docker-username=${DOCKER_USERNAME} --docker-password=${DOCKER_PASSWORD} --namespace codacy`. These credentials will be encoded inside the cluster.
+3. For codacy to work, we must set a file that which contains information like how many replicas of a given pod will codacy have, what are the cpu and memory limits for these pods, which secret tokens to use for the filestore, among others.
+   1. Define the following variables
+      1. `CODACY_PROTOCOL` - http or https protocol for your codacy installation.
+      2. `CODACY_URL` - desired URL for your codacy installation.
+      3. `CRYPTO_SECRET` - A secret...
```
andreaTP: We should probably give guidance on how to generate such secret (e.g. min length etc.)

docs/infrastructure/microk8s/install.md
joaocosta-codacy: That's a really good point. Changing it!

docs/infrastructure/microk8s/install.md
joaocosta-codacy: True. I suppose the clients can monitor the cluster after helm concludes installing.

docs/infrastructure/microk8s/install.md
joaocosta-codacy: It makes sense. I will phrase it differently such as 'the file below is an example of a valid configuration' and point them to the docs (upstream) so they know what can be configured.

docs/infrastructure/microk8s/install.md
joaocosta-codacy: Nah it can be entirely up to them. Good point! 👍

docs/infrastructure/microk8s/install.md
joaocosta-codacy: The only reason why it is in that order was just because the text seemed to flow better. First everything about microk8s, then everything helm related. I guess @pauloribeiro-codacy could help with this. :)

docs/infrastructure/microk8s/install.md
joaocosta-codacy: yes!

docs/infrastructure/microk8s/install.md
joaocosta-codacy: Good point!

docs/infrastructure/microk8s/install.md
joaocosta-codacy: @h314to might be able to shed some light onto this issue

docs/infrastructure/microk8s/install.md
joaocosta-codacy: they seem to have moved the documentation to https://microk8s.io/docs/troubleshooting#common-issues

i am replacing the link as this issue is still documented there.

docs/infrastructure/microk8s/install.md
```diff
@@ -0,0 +1,274 @@
...
+
+## Cookbook
+
+Assuming that you are starting from a blank slate, the first step is to install microk8s. Otherwise, please jump on to the [Configuring microk8s](###Configuring-microk8s) section.
+
+### Installing microk8s
+
+1. Ensure that you are working off the root folder: `cd /root`.
+2. Make sure the machine is up to date and also has the `nfs-common` package installed: `apt update && apt install nfs-common -y`.
+3. We must run configure the instance's list of hosts with localhost. This in turn will allow the correct resolution of the local microk8s cluster on the kubeconfig. We can do so by running `sed -i -e "1s/127.0.0.1 localhost.*/127.0.0.1 localhost $(hostname)/" /etc/hosts`.
```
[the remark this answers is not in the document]
joaocosta-codacy: It is EC2 instance related. I am adding some info to reflect that. @h314to would you want to add something else here?

docs/infrastructure/microk8s/install.md
```diff
@@ -0,0 +1,274 @@
...
+  * Further versions might work but we recommend using an LTS release.
+* a remote session (ssh) onto your machine of choice.
+
+## Cookbook
+
+Assuming that you are starting from a blank slate, the first step is to install microk8s. Otherwise, please jump on to the [Configuring microk8s](###Configuring-microk8s) section.
+
+### Installing microk8s
+
+1. Ensure that you are working off the root folder: `cd /root`.
```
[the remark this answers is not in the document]
joaocosta-codacy: Because it was on the script that I was looking at. :-p 
No longer required though. :)

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,273 @@
...
+   2. `sudo microk8s.enable storage && sudo microk8s.status --wait-ready`
+   3. `sudo microk8s.enable ingress &&  sudo microk8s.status --wait-ready`
+2. Now that these plugins have been enabled, we should restart microk8s to ensure a proper bootstrap of the cluster: `microk8s.stop && microk8s.start && microk8s.status --wait-ready`. Further information on this subject can be [found here](https://github.com/ubuntu/microk8s/issues/493#issuecomment-498167435).
+3. The plugins are now enabled and the cluster bootstrapped. However, we must still wait for some microk8s internals (dns, http, and ingress) plugins to be ready. Failing to do so can result in pods entering a `CrashLoopBackoff` state:
+   1. ```microk8s.kubectl wait -n kube-system --for=condition=Ready pod -l k8s-app=kube-dns --timeout=300s && microk8s.kubectl wait -n kube-system --for=condition=Ready pod -l k8s-app=hostpath-provisioner --timeout=300s && microk8s.kubectl wait -n default --for=condition=Ready pod -l app=default-http-backend --timeout=300s && microk8s.kubectl wait -n default --for=condition=Ready pod -l name=nginx-ingress-microk8s --timeout=300s```
```
paulopontesm:
```suggestion
    
   ```microk8s.kubectl wait -n kube-system --for=condition=Ready pod -l k8s-app=kube-dns --timeout=300s && \
   microk8s.kubectl wait -n kube-system --for=condition=Ready pod -l k8s-app=hostpath-provisioner --timeout=300s && \
   microk8s.kubectl wait -n default --for=condition=Ready pod -l app=default-http-backend --timeout=300s && \
   microk8s.kubectl wait -n default --for=condition=Ready pod -l name=nginx-ingress-microk8s --timeout=300s```
```

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,273 @@
...
+### 1. Helm and Tiller
+
+First we must install `helm` onto the cluster. With `helm`, we can easily install charts and manage the lifecycle of the installed artifacts. One example is the ability to rollback a failed install.
+1. Assuming you would want to install `helm` version 2.16.3, export the following environment variable: `export HELM_VERSION=2.16.3`
+   1. Please note that we currently do *not* support helm v3.
+2. Run this script that will retrieve the desired `helm` version and install it onto the machine's `/usr/local/bin`: `HELM_PKG=helm-${HELM_VERSION}-linux-amd64.tar.gz
+wget https://get.helm.sh/$HELM_PKG
+tar xvzf $HELM_PKG
+su mv linux-amd64/tiller linux-amd64/helm /usr/local/bin
+rm -rvf $HELM_PKG linux-amd64/`.
```
paulopontesm:
```suggestion
2. Run this script that will retrieve the desired `helm` version and install it onto the machine's `/usr/local/bin`: 

    ```bash
    $ HELM_VERSION=2.16.3 
    $ HELM_PKG=helm-${HELM_VERSION}-linux-amd64.tar.gz
    $ wget https://get.helm.sh/$HELM_PKG
    $ tar xvzf $HELM_PKG
    $ su mv linux-amd64/tiller linux-amd64/helm /usr/local/bin
    $ rm -rvf $HELM_PKG linux-amd64/
    ```

```

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,279 @@
...
+4. We must now deploy `tiller` onto the cluster.
+   1. First we must set a `serviceaccount` and a `clusterrolebinding` for `tiller` on the microk8s cluster: `microk8s.kubectl create serviceaccount --namespace kube-system tiller
+microk8s.kubectl create clusterrolebinding tiller-cluster-rule --clusterrole=cluster-admin --serviceaccount=kube-system:tiller`
+   2. Then we can install `tiller`, which will be installed with the same `HELM_VERSION` that you have specified earlier: `helm init --service-account tiller`
+   3. We wait for tiller to be ready: `microk8s.kubectl -n kube-system wait --for=condition=Ready pod -l name=tiller --timeout=300s`
+
+### 2. Codacy chart
```
paulopontesm: We should make use of the other docs as much as possible and put only the diffs here: https://github.com/codacy/chart/blob/master/docs/install.md

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,279 @@
...
+
+1. Create a namespace inside the cluster onto which codacy will be deployed. We use the `codacy` namespace on the commands below, but you are free to use your own value for this namespace. `microk8s.kubectl create namespace codacy`
+2. The microk8s cluster must know some docker credentials so that the container docker images can be pulled onto the pods. These credentials must belong to the namespace where codacy is installed. You can do so by running: `microk8s.kubectl create secret docker-registry docker-credentials --docker-username=${DOCKER_USERNAME} --docker-password=${DOCKER_PASSWORD} --namespace codacy`. These credentials will be encoded inside the cluster.
+3. For codacy to work, we must set a file that which contains information like how many replicas of a given pod will codacy have, what are the cpu and memory limits for these pods, which secret tokens to use for the filestore, among others.
+   1. Define the following variables
```
paulopontesm: This is very high maintenance and should already be covered in the installation docs: https://github.com/codacy/chart/blob/master/docs/install.md

docs/infrastructure/microk8s-quickstart.md
paulopontesm: this sections needs to be as little as possible

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,279 @@
...
+
+## 3. Installing microk8s
+
+You can follow the official documentation on how to [install microk8s](https://microk8s.io/docs/). However, because we were running off of an AWS EC2 instance, we had do the following steps becore installing microk8s:
+
+1. Make sure the machine is up to date and also has the `nfs-common` package installed: `apt update && apt install nfs-common -y`.
+2. We must run configure the instance's list of hosts with localhost. This was because our machine was in fact an AWS EC2 instance, and therefore we must allow the correct resolution of the local microk8s cluster on the kubeconfig. We did so by running `sed -i -e "1s/127.0.0.1 localhost.*/127.0.0.1 localhost $(hostname)/" /etc/hosts`.
+3. Traffic forwarding must be configured on the machine so that the local microk8s cluster pods can reach the internet, again because of using an AWS EC2 instance. This can be done by running `iptables -P FORWARD ACCEPT` (more info on this subject can be [found here](https://microk8s.io/docs/troubleshooting#common-issues).
```
dan-codacy: These commands require root/sudo

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,348 @@
+# Setting up a microk8s cluster
+
+Follow the steps below to set up a microk8s cluster from scratch, including all the necessary dependencies and configuration.
+
+## 1. Concepts
+
+### Kubeconfig
+
+Any tool that operates on top of a cluster needs to know which cluster it is going to work on. This execution context is typically called the `kubeconfig` which is stored in the `~/.kube/config` file. This filed basically defines a list of clusters and it holds several pieces of information per cluster entry: the address of the cluster (e.g. https://104.248.34.242:164309), its user friendly name (e.g. microk8s-cluster). It finally holds a reference to the current context. This is the active cluster against which any kubernetes control (`kubectl`) command you run will get executed.
```
pauloribeiro-codacy:
```suggestion
Any tool that operates on top of a cluster needs to know which cluster it is going to work on. This execution context is typically called the `kubeconfig` which is stored in the `~/.kube/config` file. This file basically defines a list of clusters and it holds several pieces of information per cluster entry:

* The address of the cluster (e.g. https://104.248.34.242:164309).
* The user friendly name of the cluster (e.g. microk8s-cluster).
* A reference to the current context. This is the active cluster against which any kubernetes control (`kubectl`) command you run will get executed.
```

pauloribeiro-codacy: Fix typo "filed" and use bullet list for readability.

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,348 @@
...
+
+### Helm and tiller
+
+Two executables will get installed onto the cluster as part of this process: `helm` and `tiller`. The former is responsible for resolving the configuration of the chart to be installed, while issuing the correct install commands onto the cluster. The latter is responsible for receiving the install commands issued by `helm`, as well as managing the lifecycle of the components that have been installed. `helm` is the client facing side, while `tiller` is the server/cluster facing side.
```
pauloribeiro-codacy:
```suggestion
Two executables will get installed onto the cluster as part of this process: `helm` and `tiller`.

* `helm` is responsible for resolving the configuration of the chart to be installed, while issuing the correct install commands onto the cluster.
* `tiller` is responsible for receiving the install commands issued by `helm`, as well as managing the lifecycle of the components that have been installed.

`helm` is the client facing side, while `tiller` is the server/cluster facing side.
```

pauloribeiro-codacy: Use bullet list for readability.

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,348 @@
...
+
+## 2. Prepare your environment
+
+Prepare your environment to set up the microk8s cluster:
+
+1. a machine running Ubuntu 18.04 LTS version.
+2. establish a remote session (ssh) onto this machine.
```
pauloribeiro-codacy:
```suggestion
1. A machine running Ubuntu 18.04 LTS.
2. Establish a remote SSH session onto this machine.
```

pauloribeiro-codacy: Minor style suggestions.

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,348 @@
...
+
+Assuming that you are starting from a blank slate, the first step is to install microk8s. Otherwise, please jump on to the [Configuring microk8s](###Configuring-microk8s) section.
```
pauloribeiro-codacy:
```suggestion
Assuming that you are starting from a blank slate, the first step is to install microk8s. Otherwise, please jump on to the section [Configuring microk8s](###4.-Configuring-microk8s).
```

pauloribeiro-codacy: Make sure the internal links are working on the final version.

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,348 @@
...
+
+## 3. Installing microk8s
+
+1. Make sure the machine is up to date and also has the `nfs-common` package installed
+   >apt update && apt install nfs-common -y
```
pauloribeiro-codacy:
```suggestion

   ```bash
   sudo apt update && apt install nfs-common -y
   ```
```

pauloribeiro-codacy: Consider using fenced code blocks throughout the document, specifying the language for code highlighting.

pauloribeiro-codacy: The user must run the command as super user.

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,348 @@
...
+
+2. Install microk8s
+   >sudo snap install microk8s --classic --channel=1.15/stable
```
pauloribeiro-codacy: On my Linux machine based on Ubuntu 18.04 LTS, snapd is not installed by default. I had to run:

```bash
sudo apt install snapd
```

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,348 @@
...
+   1. We recommend following the official documentation on how to [install microk8s](https://microk8s.io/docs/).
```
pauloribeiro-codacy: It would be great if we could clarify the exact steps from the microk8s documentation that must be followed. All of them? Only up to a certain point?

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,348 @@
...
+
+    >sudo microk8s.enable storage && sudo microk8s.status --wait-ready
+
+    >sudo microk8s.enable ingress &&  sudo microk8s.status --wait-ready
+
+2. Now that these plugins have been enabled, we should restart microk8s to ensure a proper bootstrap of the cluster
+    >microk8s.stop && microk8s.start && microk8s.status --wait-ready
+
+    Further information on this subject can be [found here](https://github.com/ubuntu/microk8s/issues/493#issuecomment-498167435).
+3. The plugins are now enabled and the cluster bootstrapped. However, we must still wait for some microk8s internals (dns, http, and ingress) plugins to be ready. Failing to do so can result in pods entering a `CrashLoopBackoff` state:
```
pauloribeiro-codacy:
```suggestion

3. The plugins are now enabled and the cluster bootstrapped. However, we must still wait for some microk8s internals (dns, http, and ingress) plugins to be ready. Failing to do so can result in pods entering a `CrashLoopBackoff` state:
```

pauloribeiro-codacy: Empty line between list items.

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,348 @@
...
+
+After runing these commands, we have ensured that dns, http, and nginx ingress are enabled and working properly inside the cluster.
+
+## 5. Installing Codacy
+
+### 1. Helm and Tiller
+
+First we must install `helm` onto the cluster. With `helm`, we can easily install charts and manage the lifecycle of the installed artifacts. One example is the ability to rollback a failed install.
+
+1. Assuming you would want to install `helm` version 2.16.3, export the following environment variable
```
pauloribeiro-codacy: Please see the `helm` version that we are requiring in the `index.md` file (section **Preparing to install Codacy**). We should be consistent here.

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,348 @@
...
+   1. Define a HELM_PKG variable
+      >HELM_PKG=helm-v${HELM_VERSION}-linux-amd64.tar.gz
+   2. Get the package
+      >wget https://get.helm.sh/$HELM_PKG
+   3. Extract the tarball
+      >tar xvzf $HELM_PKG
+   4. Move the binaries to `/usr/local/bin`
+      >mv linux-amd64/tiller linux-amd64/helm /usr/local/bin
+   5. Delete the extracted tarball directory
+      >rm -rvf $HELM_PKG linux-amd64/
```
pauloribeiro-codacy: Include all the commands in a fenced code block so they're easier to copy+paste at once. We can still explain the main steps of the script below the code block.

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,348 @@
...
+## 5. Installing Codacy
+
+### 1. Helm and Tiller
+
+First we must install `helm` onto the cluster. With `helm`, we can easily install charts and manage the lifecycle of the installed artifacts. One example is the ability to rollback a failed install.
+
+1. Assuming you would want to install `helm` version 2.16.3, export the following environment variable
+    >export HELM_VERSION=2.16.3
+   1. Please note that we currently do *not* support helm v3.
+2. Run this script that will retrieve the desired `helm` version and install it onto the machine's `/usr/local/bin`:
```
pauloribeiro-codacy: Notice that in the `index.md` we are already asking the user to install the required tools by following their documentation. So this step is probably not necessary?

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,348 @@
...
+
+3. We must now deploy `tiller` onto the cluster.
+   1. First we must set a `serviceaccount` and a `clusterrolebinding` for `tiller` on the microk8s cluster:
+      >microk8s.kubectl create serviceaccount --namespace kube-system tiller
+
+      >microk8s.kubectl create clusterrolebinding tiller-cluster-rule --clusterrole=cluster-admin --serviceaccount=kube-system:tiller
+   1. Then we can install `tiller`, which will be installed with the same `HELM_VERSION` that you have specified earlier:
+      >helm init --service-account tiller
+   1. We wait for tiller to be ready
+      >microk8s.kubectl -n kube-system wait --for=condition=Ready pod -l name=tiller --timeout=300s
```
pauloribeiro-codacy: Same recommendation as above, condense all the commands in a single fenced code block and explain what this does below the code block.

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,348 @@
...
+    >microk8s.kubectl wait -n ingress --for=condition=Ready pod -l name=nginx-ingress-microk8s --timeout=300s
+
+4. Export your microk8s cluster configuration to the local kubeconfig file
+    >microk8s.config > ~/.kube/config
+
+After runing these commands, we have ensured that dns, http, and nginx ingress are enabled and working properly inside the cluster.
+
+## 5. Installing Codacy
+
+### 1. Helm and Tiller
```
pauloribeiro-codacy: I would move this section to **Configuring microk8s**, as we're still taking care of the prerequisites and not installing Codacy yet.

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,348 @@
...
+
+### 2. Codacy chart
+
+Please refer to [our chart installation documentation](https://github.com/codacy/chart/blob/master/docs/install.md).
+
+Additionally, any `kubectl` command from [our chart installation](https://github.com/codacy/chart/blob/master/docs/install.md) must be executed as a `microk8s.kubectl` command.
+
+#### 1. Default Values
+
+We provide a base (values file)[##microk8s-values.yaml] that you can use for your microk8s Codacy installation. This file provisions a smaller number of component replicas with limited resources (i.e. CPU and memory). Should you want to use this as a template, please pass your `microk8s-values.yaml` file on (4.) of [our chart installation](https://github.com/codacy/chart/blob/master/docs/install.md). Lastly, it is important to correctly set your `codacy-api.ingress.hosts.host` if you are using this file as a template.
```
pauloribeiro-codacy: Consider providing the `.yaml` file as a standalone file in the GitHub repository just like we did for AKS and EKS:

https://github.com/codacy/chart/tree/master/docs/infrastructure/AKS
https://github.com/codacy/chart/tree/master/docs/infrastructure/EKS

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,348 @@
...
+
+#### 2. External Databases
+
+If you have decided to go with the [default values for the installation](####1.Default-Values), your microk8s cluster will contain a postgresql database within itself. This is not recommended as the data stored in the database will not be as durable if you store it in an external database.
+To use an external database, you should follow the following steps:
```
pauloribeiro-codacy:
```suggestion
If you have decided to go with the [default values for the installation](####1.-Default-Values), your microk8s cluster will contain a PostgreSQL database within itself. This is not recommended as the data stored in the database will not be as durable if you store it in an external database.

Follow the steps below to use an external database:
```

pauloribeiro-codacy: Minor style adjustments.

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,348 @@
...
+
+Please refer to [our chart installation documentation](https://github.com/codacy/chart/blob/master/docs/install.md).
+
+Additionally, any `kubectl` command from [our chart installation](https://github.com/codacy/chart/blob/master/docs/install.md) must be executed as a `microk8s.kubectl` command.
+
+#### 1. Default Values
+
+We provide a base (values file)[##microk8s-values.yaml] that you can use for your microk8s Codacy installation. This file provisions a smaller number of component replicas with limited resources (i.e. CPU and memory). Should you want to use this as a template, please pass your `microk8s-values.yaml` file on (4.) of [our chart installation](https://github.com/codacy/chart/blob/master/docs/install.md). Lastly, it is important to correctly set your `codacy-api.ingress.hosts.host` if you are using this file as a template.
+
+#### 2. External Databases
```
pauloribeiro-codacy: Here we already have instructions on how to create the external database:

https://github.com/codacy/chart/blob/master/docs/requirements.md#postgresql-server-setup

Are there steps that are significantly different when using microk8s? If not, we can assume that the user already has an external database as per the requirements.

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,348 @@
...
+    CREATE DATABASE results WITH OWNER=<user>;
+    CREATE DATABASE metrics WITH OWNER=<user>;
+    CREATE DATABASE filestore WITH OWNER=<user>;
+    CREATE DATABASE jobs WITH OWNER=<user>;
+    CREATE DATABASE activities WITH OWNER=<user>;
+    CREATE DATABASE hotspots WITH OWNER=<user>;
+    CREATE DATABASE listener WITH OWNER=<user>;
+    ```
+
+## 6. Accessing the Codacy UI
```
pauloribeiro-codacy: This should be tied better with this section:

https://github.com/codacy/chart/blob/master/docs/index.md#post-install-configuration

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,348 @@
...
+1. We had to configure the instance's list of hosts with localhost. This was because our machine was in fact an AWS EC2 instance, and therefore we must allow the correct resolution of the local microk8s cluster on the kubeconfig. We did so by running
+
+    >sudo sed -i -e "1s/127.0.0.1 localhost.*/127.0.0.1 localhost $(hostname)/" /etc/hosts
+
+2. Traffic forwarding must be configured on the machine so that the local microk8s cluster pods can reach the internet, again because of using an AWS EC2 instance. This can be done by running
+    >sudo iptables -P FORWARD ACCEPT
+
+    (more info on this subject can be [found here](https://microk8s.io/docs/troubleshooting#common-issues)).
+
+## microk8s-values.yaml
```
pauloribeiro-codacy: See comment above about providing this as a standalone file. :)

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,348 @@
...
+    CREATE DATABASE listener WITH OWNER=<user>;
+    ```
+
+## 6. Accessing the Codacy UI
+
+In the previous section, you should have defined values for `codacy-api.ingress.hosts.host`, `global.codacy.url`, and `global.codacy.backendUrl`. This value will be the address where you can access the Codacy UI from your web browser.
+
+## 7. Troubleshooting
+
+### 1. AWS EC2 specific issues
```
pauloribeiro-codacy: I think this may be quite specific to the test that you did and may confuse customers since we also provide instructions to install Codacy on Amazon EKS.

I was thinking that customers using microk8s will use any kind of infrastructure, most probably one that is managed by them?

Consider removing this section.

docs/infrastructure/microk8s-quickstart.md
joaocosta-codacy: The only thing missing is to mention in the docs to add the specific values to the values file i guess. To have something like an example block. Let me iterate on this. :)

docs/infrastructure/microk8s-quickstart.md
joaocosta-codacy: I am not 100% sure what to do about it :/ suggestions?

docs/infrastructure/microk8s-quickstart.md
joaocosta-codacy: 👍 removing

docs/infrastructure/microk8s-quickstart.md
pauloribeiro-codacy: Sorry, I confused the files. The ones in the link above are the Terraform backend files and have nothing to do with the `microk8s-values.yaml` file mentioned here.

The file I wanted to mention is actually this one:
https://github.com/codacy/chart/blob/master/codacy/values-production.yaml

docs/infrastructure/microk8s-quickstart.md
pauloribeiro-codacy: I confused the files in my comment above, sorry.

What I meant is that we are providing the following file in the GitHub repository itself:

https://github.com/codacy/chart/blob/master/codacy/values-production.yaml

We could also provide this `microk8s-values.yaml` in the same way so that it's a bit easier for users to download and use it - and it also makes this documentation shorter.

I suggest calling the file `values-production-microk8s.yaml` so it is ordered alphabetically and stays close to the `values-production.yaml` one.

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,325 @@
...
+2. Establish a remote SSH session onto this machine.
+
+Assuming that you are starting from a blank slate, the first step is to install microk8s. Otherwise, please jump on to the section [Configuring microk8s](###4.-Configuring-microk8s).
+
+## 3. Installing microk8s
+
+1. Make sure the machine is up to date and also has the `nfs-common` package installed
+
+   ```bash
+   sudo apt update && apt install nfs-common -y
```
pauloribeiro-codacy: Must close the fenced code block:

```suggestion
   ```bash
   sudo apt update && apt install nfs-common -y
   ```
```

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,325 @@
...
+* `tiller` is responsible for receiving the install commands issued by `helm`, as well as managing the lifecycle of the components that have been installed.
+
+`helm` is the client facing side, while `tiller` is the server/cluster facing side.
+
+## 2. Prepare your environment
+
+Prepare your environment to set up the microk8s cluster:
+
+1. A machine running Ubuntu 18.04 LTS.
+2. Establish a remote SSH session onto this machine.
```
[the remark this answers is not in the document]
pauloribeiro-codacy: Yes, I agree!

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,325 @@
...
+Assuming that you are starting from a blank slate, the first step is to install microk8s. Otherwise, please jump on to the section [Configuring microk8s](###4.-Configuring-microk8s).
+
+## 3. Installing microk8s
+
+1. Make sure the machine is up to date and also has the `nfs-common` package installed
+
+   ```bash
+   sudo apt update && apt install nfs-common -y
+
+2. Install microk8s
```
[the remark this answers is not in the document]
pauloribeiro-codacy: I like this idea. As it is now, it seems that the user must do something "Codacy-specific" to install microk8s and then also follow the microk8s installation documentation.

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,325 @@
...
+
+After runing these commands, we have ensured that dns, http, and nginx ingress are enabled and working properly inside the cluster.
+
+## 5. Installing Codacy
+
+### 1. Helm and Tiller
+
+First we must install `helm` onto the cluster. With `helm`, we can easily install charts and manage the lifecycle of the installed artifacts. One example is the ability to rollback a failed install.
+
+1. Assuming you would want to install `helm` version 2.16.3, export the following environment variable
```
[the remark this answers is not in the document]
pauloribeiro-codacy: Also, this is already described here:

https://github.com/codacy/chart/blob/master/docs/index.md#preparing-to-install-codacy

docs/infrastructure/microk8s-quickstart.md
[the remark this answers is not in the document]
joaocosta-codacy: 👍

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,325 @@
...
+`helm` is the client facing side, while `tiller` is the server/cluster facing side.
+
+## 2. Prepare your environment
+
+Prepare your environment to set up the microk8s cluster:
+
+1. A machine running Ubuntu 18.04 LTS.
+2. Establish a remote SSH session onto this machine.
+
+Assuming that you are starting from a blank slate, the first step is to install microk8s. Otherwise, please jump on to the section [Configuring microk8s](###4.-Configuring-microk8s).
```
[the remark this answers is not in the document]
joaocosta-codacy: 👍

docs/infrastructure/microk8s-quickstart.md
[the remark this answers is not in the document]
joaocosta-codacy: The thing is they need to install up to 1.15 microk8s, which translates to kubernetes 1.15. Any version above that (like the latest 1.17) will fail installing codacy.

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,325 @@
...
+   5. Delete the extracted tarball directory
+      >rm -rvf $HELM_PKG linux-amd64/
+
+3. We must now deploy `tiller` onto the cluster.
+   1. First we must set a `serviceaccount` and a `clusterrolebinding` for `tiller` on the microk8s cluster:
+      >microk8s.kubectl create serviceaccount --namespace kube-system tiller
+
+      >microk8s.kubectl create clusterrolebinding tiller-cluster-rule --clusterrole=cluster-admin --serviceaccount=kube-system:tiller
+   1. Then we can install `tiller`, which will be installed with the same `HELM_VERSION` that you have specified earlier:
+      >helm init --service-account tiller
```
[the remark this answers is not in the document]
joaocosta-codacy: We only support up to helm 2.16.3 (which is the latest helm2)
I think i need to mention that helm version here then.

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,325 @@
...
+The crucial configuration step is that you must have the codacy-api ingress and backend urls(`codacy-api.ingress.hosts.host`, `global.codacy.url`, and `global.codacy.backendUrl`) properly configured.
+
+## microk8s-values.yaml
+
+```yaml
+global:
+  imagePullSecrets:
+    - name: docker-credentials
+  codacy:
+    url: "${CODACY_PROTOCOL}://${CODACY_URL}"
```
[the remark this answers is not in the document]
joaocosta-codacy: Good point 👍 you should write them to the file. Fixing this.

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,325 @@
...
+
+#### 1. Default Values
+
+We provide a base (values file)[##microk8s-values.yaml] that you can use for your microk8s Codacy installation. This file provisions a smaller number of component replicas with limited resources (i.e. CPU and memory). Should you want to use this as a template, please pass your `microk8s-values.yaml` file on (4.) of [our chart installation](https://github.com/codacy/chart/blob/master/docs/install.md). Lastly, it is important to correctly set your `codacy-api.ingress.hosts.host` if you are using this file as a template.
+
+#### 2. External Databases
+
+If you have decided to go with the [default values for the installation](####1.-Default-Values), your microk8s cluster will contain a PostgreSQL database within itself. This is not recommended as the data stored in the database will not be as durable if you store it in an external database.
+
+To setup an external database, you should follow the [steps described here](../requirements.md).
```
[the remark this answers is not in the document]
joaocosta-codacy: I will leave that call for @pauloribeiro-codacy to make :) Just because we don't want to repeat ourselves across the docs so maybe we need to add that info elsewhere (not on microk8s docs)

docs/infrastructure/microk8s-quickstart.md
andreaTP: super important, we do not support k8s > 16, so you should install microk8s  at a very specific version. cc. @h314to for the help

docs/infrastructure/microk8s-quickstart.md
joaocosta-codacy: Good idea. I wouldn't call it 'production' however. It is mostly just a placeholder/template/starting point. But i have moved it to live alongside the other values files :)

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,325 @@
...
+    >microk8s.stop && microk8s.start && microk8s.status --wait-ready
+
+    Further information on this subject can be [found here](https://github.com/ubuntu/microk8s/issues/493#issuecomment-498167435).
+3. The plugins are now enabled and the cluster bootstrapped. However, we must still wait for some microk8s internals (dns, http, and ingress) plugins to be ready. Failing to do so can result in pods entering a `CrashLoopBackoff` state:
+
+    >microk8s.kubectl wait -n kube-system --for=condition=Ready pod -l k8s-app=kube-dns --timeout=300s
+
+    >microk8s.kubectl wait -n kube-system --for=condition=Ready pod -l k8s-app=hostpath-provisioner --timeout=300s
+
+    >microk8s.kubectl wait -n ingress --for=condition=Ready pod -l name=nginx-ingress-microk8s --timeout=300s
```
andreaTP: this command should be:
```
microk8s.kubectl wait -n default --for=condition=Ready pod -l name=nginx-ingress-microk8s
```

andreaTP:
the command:
```
helm upgrade --install codacy codacy-stable/codacy \
  --namespace codacy \
  --recreate-pods \
  --values values.yaml
```
here: https://github.com/codacy/chart/blob/master/docs/install.md
should have the `--atomic` option

docs/infrastructure/microk8s-quickstart.md
joaocosta-codacy: References in Installing 3.2

docs/infrastructure/microk8s-quickstart.md
joaocosta-codacy: But helm is not microk8s specific. It is required to install the codacy chart :/

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,325 @@
...
+
+4. Export your microk8s cluster configuration to the local kubeconfig file
```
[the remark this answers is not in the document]
joaocosta-codacy: not exactly required, you are right. I used it so that i could use k9s for instance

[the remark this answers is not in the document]
joaocosta-codacy: removing

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,325 @@
...
+
+    >sudo microk8s.enable ingress &&  sudo microk8s.status --wait-ready
+
+2. Now that these plugins have been enabled, we should restart microk8s to ensure a proper bootstrap of the cluster
+    >microk8s.stop && microk8s.start && microk8s.status --wait-ready
+
+    Further information on this subject can be [found here](https://github.com/ubuntu/microk8s/issues/493#issuecomment-498167435).
+3. The plugins are now enabled and the cluster bootstrapped. However, we must still wait for some microk8s internals (dns, http, and ingress) plugins to be ready. Failing to do so can result in pods entering a `CrashLoopBackoff` state:
+
+    >microk8s.kubectl wait -n kube-system --for=condition=Ready pod -l k8s-app=kube-dns --timeout=300s
```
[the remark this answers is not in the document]
joaocosta-codacy: I don't think so because one is waiting for kube-system, and the other is waiting microk8s itself :/
as for the timeouts, you are absolutely right. removing them :)

docs/infrastructure/microk8s-quickstart.md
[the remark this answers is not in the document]
pauloribeiro-codacy: The initial version of the EKS documentation also mentioned a "built-in" database and then making the installation production-ready with an external one. But due to feedback we decided to simplify and include explicit instructions only on how to set up Codacy with the (recommended) external database. So, I suggest that we focus on the recommended path of setting up the connection to the external database that is already part of the pre-requirements. You can check with @paulopontesm regarding what was done for EKS.

andreaTP:
I thought it was here but I don't see it anymore:
```
sudo echo "--allow-privileged=true" >> /var/snap/microk8s/current/args/kube-apiserver
```
is missing

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,325 @@
...
+  replicaCount: 1
+  resources:
+    limits:
+      cpu: 1
+      memory: 2Gi
+    requests:
+      cpu: 0.5
+      memory: 1Gi
+
+postgres:
```
andreaTP: cc. @h314to is this an option we want to support?

joaocosta-codacy:
> I thought it was here but I don't see it anymore:
> 
> ```
> sudo echo "--allow-privileged=true" >> /var/snap/microk8s/current/args/kube-apiserver
> ```
> 
> is missing

Where should this go? It is not pinned to any file/line :/

codacy/values-microk8s.yaml
```diff
@@ -0,0 +1,184 @@
...
+
+listener:
+  replicaCount: 1
+  resources:
+    limits:
+      cpu: 2
+      memory: 8Gi
+    requests:
+      cpu: 1
+      memory: 6Gi
```
andreaTP: those values are not in-line with latest agreement, correct?

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,138 @@
...
+   sudo apt update && apt install nfs-common -y
+
+2. Install microk8s. Codacy is supported up to kubernetes 1.15. Therefore we recommend following the [official documentation on how to install microk8s](https://microk8s.io/docs/) with the `1.15/stable` channel.
+
+## 4. Configuring microk8s
+
+1. First, we must enable the following plugins on microk8s:
+
+    ```bash
+    microk8s.enable dns && microk8s.status --wait-ready
```
andreaTP: the `microk8s.status --wait-ready` is redundant since it's performed at the end

docs/install.md
```diff
@@ -40,7 +40,7 @@ Follow the steps below to install Codacy on an existing Kubernetes cluster using
     helm repo update
     helm upgrade --install codacy codacy-stable/codacy \
       --namespace codacy \
-      --recreate-pods \
+      --atomic \
```
andreaTP: we need to set a higher timeout as well here... I guess ...

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,138 @@
...
+* `tiller` is responsible for receiving the install commands issued by `helm`, as well as managing the lifecycle of the components that have been installed.
+
+`helm` is the client facing side, while `tiller` is the server/cluster facing side.
+
+## 2. Prepare your environment
+
+Prepare your environment to set up the microk8s cluster:
+
+1. A machine running Ubuntu 18.04 LTS.
+2. Establish a session (local or remote) onto this machine.
```
paulopontesm: An RDS and firewall open (maybe a direct link from here to the postgres requirements)

docs/install.md
joaocosta-codacy: true. Does adding a timeout of 5 min sounds reasonable?

codacy/values-microk8s.yaml
joaocosta-codacy: True, they are not. I am going to update the file.

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,148 @@
...
+
+Additionally, any `kubectl` command from [our chart installation](https://github.com/codacy/chart/blob/master/docs/install.md) must be executed as a `microk8s.kubectl` command.
+
+#### 1. Default Values
+
+We provide a base (values file)[##values-microk8s.yaml] that you can use for your microk8s Codacy installation. This file provisions a smaller number of component replicas with limited resources (i.e. CPU and memory). Should you want to use this as a template, please pass your `values-microk8s.yaml` file on (4.) of [our chart installation](https://github.com/codacy/chart/blob/master/docs/install.md). Lastly, it is important to correctly set your `codacy-api.ingress.hosts.host` if you are using this file as a template.
+
+#### 2. External Databases
+
+If you have decided to go with the [default values for the installation](####1.-Default-Values), your microk8s cluster will contain a PostgreSQL database within itself. This is not recommended as the data stored in the database will not be as durable if you store it in an external database.
```
h314to: We really should not make it easy for people to install using the internal postgres, because it will end up causing them lots of problems

codacy/values-microk8s.yaml
```diff
@@ -0,0 +1,184 @@
...
+      cpu: 2000m
+      memory: 15000Mi
+    requests:
+      cpu: 1000m
+      memory: 10000Mi
+
+worker-manager:
+  replicaCount: 1
+  resources:
+    limits:
```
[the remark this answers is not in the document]
joaocosta-codacy: I asume it is safe to do that now?

docs/infrastructure/microk8s-quickstart.md
joaocosta-codacy: I agree. I am removing this paragraph.

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,136 @@
...
+1. A machine running Ubuntu 18.04 LTS.
+2. A postgres database and open firewall rule. ([Documentation can be found here.](../requirements.md))
+
+To start installing Codacy, you will need to:
+
+1. Establish a session (local or remote) onto the machine described in 1. above.
+
+Assuming that you are starting from a blank slate, the first step is to install microk8s. Otherwise, consider that Codacy is supported up to kubernetes 1.15. Please jump on to the section [Configuring microk8s](###4.-Configuring-microk8s) if you are on a compatible kubernetes version.
+
+## 3. Installing microk8s
```
paulopontesm: suggestion from rodrigo:
Single script with verification at the begging if we are on the right OS, kernel, etc

codacy/values-microk8s.yaml
```diff
@@ -0,0 +1,184 @@
...
+    hosts:
+    - host: '<--- codacy url --->'
+      paths:
+        - /
+
+portal:
+  replicaCount: 1
+  resources:
+    limits:
+      cpu: 500m
```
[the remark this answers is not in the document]
andreaTP: yes, please! But we can probably do it in a separate PR.

codacy/values-microk8s.yaml
```diff
@@ -0,0 +1,266 @@
...
+    postgresqlUsername: <--- db username --->
+    postgresqlDatabase: <--- db name --->
+    postgresqlPassword: <--- db password --->
+    host: <--- db url --->
+    service:
+      port: <--- db port --->
+  persistence:
+    claim:
+      size: 50Gi
+  nfsserverprovisioner:
```
andreaTP: we are missing to set requests and limits here, this is quite important:
```
  nfsserverprovisioner:
    enabled: true
    persistence:
      enabled: true
      size: 100Gi
    resources:
      limits:
        cpu: 250m
        memory: 200Mi
      requests:
        cpu: 100m
        memory: 200Mi
```
is my current local configuration, I don't know if this has impact on the analysis or not.

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,136 @@
...
+    Further information on this subject can be [found here](https://github.com/ubuntu/microk8s/issues/493#issuecomment-498167435).
+
+3. The plugins are now enabled and the cluster bootstrapped. However, we must still wait for some microk8s internals (dns, http, and ingress) plugins to be ready. Failing to do so can result in pods entering a `CrashLoopBackoff` state:
+
+   ```bash
+    microk8s.kubectl wait -n kube-system --for=condition=Ready pod -l k8s-app=kube-dns
+
+    microk8s.kubectl wait -n kube-system --for=condition=Ready pod -l k8s-app=hostpath-provisioner
+
+    microk8s.kubectl wait -n default --for=condition=Ready pod -l name=nginx-ingress-microk8s
```
jllopes: The controller is created in the ingress namespace, update this
```suggestion
    microk8s.kubectl wait -n ingress --for=condition=Ready pod -l name=nginx-ingress-microk8s
```

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,136 @@
...
+Assuming that you are starting from a blank slate, the first step is to install microk8s. Otherwise, consider that Codacy is supported up to kubernetes 1.15. Please jump on to the section [Configuring microk8s](###4.-Configuring-microk8s) if you are on a compatible kubernetes version.
+
+## 3. Installing microk8s
+
+1. Make sure the machine is up to date and also has the `nfs-common` package installed
+
+   ```bash
+   sudo apt update && apt install nfs-common -y
+
+2. Install microk8s. Codacy is supported up to kubernetes 1.15. Therefore we recommend following the [official documentation on how to install microk8s](https://microk8s.io/docs/) with the `1.15/stable` channel.
```
paulopontesm: We need to freeze also the version of the microk8s itself

docs/infrastructure/microk8s-quickstart.md
paulopontesm: I agree with @pauloribeiro-codacy . This looks a lot like setting up prerequisites to then install codacy. Microk8s is a prerequisite and Helm is another.

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,136 @@
...
+
+Additionally, any `kubectl` command from [our chart installation](https://github.com/codacy/chart/blob/master/docs/install.md) must be executed as a `microk8s.kubectl` command.
+
+#### 1. Default Values
+
+We provide a base [values file](##values-microk8s.yaml) that you can use for your microk8s Codacy installation. This file provisions a smaller number of component replicas with limited resources (i.e. CPU and memory). Should you want to use this as a template, please pass your `values-microk8s.yaml` file on (4.) of [our chart installation](https://github.com/codacy/chart/blob/master/docs/install.md). Lastly, it is important to correctly set your `codacy-api.ingress.hosts.host` if you are using this file as a template.
+
+#### 2. External Databases
+
+If you do not have an external database set up for you by an Ops team,  you should follow the [steps described here](../requirements.md).
```
jllopes: There's an extra space here
```suggestion
If you do not have an external database set up for you by an Ops team, you should follow the [steps described here](../requirements.md).
```

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,136 @@
...
+
+   ```bash
+   sudo apt update && apt install nfs-common -y
+
+2. Install microk8s. Codacy is supported up to kubernetes 1.15. Therefore we recommend following the [official documentation on how to install microk8s](https://microk8s.io/docs/) with the `1.15/stable` channel.
+
+3. Allow privileged containers for microk8s
+
+   ```bash
+   sudo echo "--allow-privileged=true" >> /var/snap/microk8s/current/args/kube-apiserver
```
jllopes: This step gave me permission denied, even with sudo. Might've been something to do with the instance I'm using, but maybe we should take a look.

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,136 @@
...
+   ```
+
+## 4. Configuring microk8s
+
+1. First, we must enable the following plugins on microk8s:
```
jllopes: These steps take a bit of time, maybe it's worth it to mention it somewhere?

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,136 @@
...
+    ```bash
+   microk8s.stop && microk8s.start && microk8s.status --wait-ready
+    ```
+
+    Further information on this subject can be [found here](https://github.com/ubuntu/microk8s/issues/493#issuecomment-498167435).
+
+3. The plugins are now enabled and the cluster bootstrapped. However, we must still wait for some microk8s internals (dns, http, and ingress) plugins to be ready. Failing to do so can result in pods entering a `CrashLoopBackoff` state:
+
+   ```bash
+    microk8s.kubectl wait -n kube-system --for=condition=Ready pod -l k8s-app=kube-dns
```
jllopes: From microk8s documentation, they actually suggest to use an alias for kubectl `alias kubectl='microk8s.kubectl'`.
Do you think we should mention it here?

docs/infrastructure/microk8s-quickstart.md
jllopes: Please mention this version before the link to the microk8s repo at the top. I followed the quickstart they have there and had to redo everything

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,136 @@
...
+
+After these commands return successfully, we have ensured that dns, http, and nginx ingress are enabled and working properly inside the cluster.
+
+## 5. Installing Codacy
+
+### 1. Helm and Tiller
+
+First we must install `helm` onto the cluster. With `helm`, we can easily install charts and manage the lifecycle of the installed artifacts. One example is the ability to rollback a failed install.
+
+1. __Codacy supports up to `helm` version 2.16.3.__
```
jllopes: Should we add the command needed to install with this specific version?

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,136 @@
...
+
+      ```bash
+      microk8s.kubectl create serviceaccount --namespace kube-system tiller
+      microk8s.kubectl create clusterrolebinding tiller-cluster-rule --clusterrole=cluster-admin --serviceaccount=kube-system:tiller
+      ```
+
+   2. Then we can install `tiller`, which will be installed with the same `HELM_VERSION` that you have specified earlier and we wait for tiller to be ready:
+
+      ```bash
+      helm init --service-account tiller
```
jllopes: I was having some troubles with this and had to do
`mkdir ~/.kube && sudo chown -Rv ubuntu:ubuntu ~/.kube && sudo microk8s.config > .kube/config`
before
`helm init --service-account tiller`
to make it work

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,136 @@
...
+      microk8s.kubectl -n kube-system wait --for=condition=Ready pod -l name=tiller
+      ```
+
+### 2. Codacy chart
+
+Please refer to [our chart installation documentation](https://github.com/codacy/chart/blob/master/docs/install.md).
+
+Additionally, any `kubectl` command from [our chart installation](https://github.com/codacy/chart/blob/master/docs/install.md) must be executed as a `microk8s.kubectl` command.
```
jllopes: As I said above, from microk8s documentation, they actually suggest to use an alias for kubectl commands, consider adding that here

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,136 @@
...
+
+#### 1. Default Values
+
+We provide a base [values file](##values-microk8s.yaml) that you can use for your microk8s Codacy installation. This file provisions a smaller number of component replicas with limited resources (i.e. CPU and memory). Should you want to use this as a template, please pass your `values-microk8s.yaml` file on (4.) of [our chart installation](https://github.com/codacy/chart/blob/master/docs/install.md). Lastly, it is important to correctly set your `codacy-api.ingress.hosts.host` if you are using this file as a template.
```
jllopes: This is a bit weird, we refer to the chart installation documentation and then we say we should use `values-microk8s.yaml`, we should mention that we need to use this one instead of `values-production.yaml`

docs/infrastructure/microk8s-quickstart.md
jllopes: Maybe we should add the command needed to do it?

docs/infrastructure/microk8s-quickstart.md
pauloribeiro-codacy: Installing Helm as part of preparing the environment to install Codacy is already mentioned here:

https://github.com/codacy/chart/blob/master/docs/index.md#preparing-to-install-codacy

We need to integrate the quickstart for the microk8s better with the main installation process in that page, though.

docs/infrastructure/microk8s-quickstart.md
pauloribeiro-codacy: My opinion: it would add an extra step / instruction and the benefits of calling a shorter command could be outweighed by the issue of collision for customers who happen to also use the "standard" kubectl.

docs/infrastructure/microk8s-quickstart.md
pauloribeiro-codacy: This has to do with the fact that the (global) directories under /var/snap are actually read-only mount points (see https://askubuntu.com/a/1035241).

So, we should have another way of doing this configuration, otherwise the workaround will be rather complex. Just an idea, maybe the snap could be installed in the user space instead, which would allow changing the script (and could we enforce the customers to do this)?...

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,112 @@
...
+    microk8s.status --wait-ready
+    ```
+
+3.  Install the helm binary
+
+    ```bash
+    HELM_PKG=helm-v2.16.3-linux-amd64.tar.gz
+    wget https://get.helm.sh/$HELM_PKG
+    tar xvzf $HELM_PKG
+    mv linux-amd64/tiller linux-amd64/helm /usr/local/bin
```
paulopontesm: confirm if this needs sudo.. according to andrea it does

docs/infrastructure/microk8s-quickstart.md
paulopontesm: Done

docs/infrastructure/microk8s-quickstart.md
paulopontesm: done

docs/infrastructure/microk8s-quickstart.md
paulopontesm: The command on installing microk8s should solve this `sudo usermod -a -G microk8s $USER `

docs/infrastructure/microk8s-quickstart.md
pauloribeiro-codacy: It does, so that you can write on the folder /usr/local/bin.

docs/infrastructure/microk8s-quickstart.md
pauloribeiro-codacy: On the other hand, it might simplify the process of following [these instructions](https://github.com/codacy/chart/blob/master/docs/install.md) with microk8s. We'll consider this idea.

docs/requirements.md
```diff
@@ -107,6 +107,7 @@ You must manually create the databases required by Codacy on the PostgreSQL serv
     CREATE DATABASE activities WITH OWNER=codacy;
     CREATE DATABASE hotspots WITH OWNER=codacy;
     CREATE DATABASE listener WITH OWNER=codacy;
+    CREATE DATABASE crow WITH OWNER=codacy;
```
joaocosta-codacy: Good catch! I had forgotten about the necessity of having this db now that crow is part of self-hosted too!
👍

docs/infrastructure/microk8s-quickstart.md
paulopontesm: Done

docs/infrastructure/microk8s-quickstart.md
```diff
@@ -0,0 +1,104 @@
...
+    sudo su - $USER
+    ```
+
+    Check that microk8s is running.
+
+    ```bash
+    microk8s.status --wait-ready
+    ```
+
+3.  Install the version `v2.16.3` helm binary
```
andreaTP:
```suggestion
3.  Install the version `v2.16.3` of the helm binary
```

codacy/values-production.yaml
```diff
@@ -197,54 +219,68 @@ core:
...
       cpu: 500m
       memory: 1000Mi
     requests:
-      cpu: 100m
-      memory: 500Mi
+      cpu: 200m
+      memory: 200Mi
   config:
     workers:
-      genericMax: 10
```
andreaTP: is this deliberate?

paulopontesm: Yes, I think we realized that 10 is clearly too much. Also, 2 is the default for doplicated. We better go with that.

PR #293, opened 2020-03-10
Improve Ingress configuration documentation

pauloribeiro-codacy:
I tried to clarify what are the actual values that the user must use to configure Ingress - hostnames of the Codacy application and API endpoint instead of "DNS" entries.

### DIFF
--- a/codacy/values.yaml
+++ b/codacy/values.yaml
@@ -263,14 +263,10 @@ codacy-ingress:
   create: true
   ingress:
     hostname:
-      app: <codacy-app.dns.internal>
-      api: <codacy-api.dns.internal>
+      app: "codacy.example.com"
+      api: "codacy.example.com"
     annotations:
       kubernetes.io/ingress.class: nginx
-      external-dns.alpha.kubernetes.io/hostname: <codacy-app.dns.internal>, <codacy-api.dns.internal>
-      external-dns.alpha.kubernetes.io/scope: public
-      nginx.ingress.kubernetes.io/proxy-body-size: 128m
-      nginx.ingress.kubernetes.io/ssl-redirect: "true"
       nginx.ingress.kubernetes.io/use-regex: "true"
     extraPaths:
       - backend:
--- a/docs/install.md
+++ b/docs/install.md
@@ -23,7 +23,7 @@ Before you continue, make sure you have the following tools installed in your ma
    kubectl create secret docker-registry docker-credentials --docker-username=$DOCKER_USERNAME --docker-password=$DOCKER_PASSWORD --namespace codacy
    ```
 
-3. Use a text editor of your choice edit the [`values-production.yaml`](../codacy/values-production.yaml) file, changing the values with placeholders as described in the comments.
+3. Use a text editor of your choice edit the [`values-production.yaml`](https://github.com/codacy/chart/blob/master/codacy/values-production.yaml) file, changing the values with placeholders as described in the comments.
 
 4. Create a record in your DNS provider with the hostname you used in the previous step to send traffic to your ingress controller.
 
@@ -33,7 +33,7 @@ Before you continue, make sure you have the following tools installed in your ma
 5. Add Codacy's chart repository to your helm client and install the Codacy chart using the values file created in the previous step.
 
    **NOTE:**
-   If you are in microk8s, don't forget to use the [`values-microk8s.yaml`](../codacy/values-microk8s.yaml) configuration file as stated [here](infrastructure/microk8s-quickstart.md#5-installing-codacy).
+   If you are in microk8s, don't forget to use the [`values-microk8s.yaml`](https://github.com/codacy/chart/blob/master/codacy/values-microk8s.yaml) configuration file as stated [here](infrastructure/microk8s-quickstart.md#5-installing-codacy).
 
    ```bash
    helm repo add codacy-stable https://charts.codacy.com/stable/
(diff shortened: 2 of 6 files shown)


codacy/values-production.yaml
```diff
@@ -66,22 +66,22 @@ codacy-ingress:
...
-      api: <--- codacy-api.dns.internal ---> # Codacy API DNS
+      app: <--- codacy-app.dns.internal ---> # Codacy application hostname
+      api: <--- codacy-api.dns.internal ---> # Codacy API hostname
     annotations:
       kubernetes.io/ingress.class: nginx
       kubernetes.io/tls-acme: "true"
       certmanager.k8s.io/acme-challenge-type: http01
       certmanager.k8s.io/cluster-issuer: letsencrypt
-      external-dns.alpha.kubernetes.io/hostname: '<--- codacy-app.dns.internal --->, <--- codacy-api.dns.internal --->' # Codacy application and API DNS 
+      external-dns.alpha.kubernetes.io/hostname: '<--- codacy-app.dns.internal --->, <--- codacy-api.dns.internal --->' # Codacy application and API hostnames
```
pauloribeiro-codacy: Is "hostnames" the correct term here? Or should it be "domain names"?

codacy/values-production.yaml
```diff
@@ -66,22 +66,22 @@ codacy-ingress:
...
     hostname:
-      app: <--- codacy-app.dns.internal ---> # Codacy application DNS 
-      api: <--- codacy-api.dns.internal ---> # Codacy API DNS
+      app: <--- codacy-app.dns.internal ---> # Codacy application hostname
+      api: <--- codacy-api.dns.internal ---> # Codacy API hostname
     annotations:
       kubernetes.io/ingress.class: nginx
       kubernetes.io/tls-acme: "true"
       certmanager.k8s.io/acme-challenge-type: http01
       certmanager.k8s.io/cluster-issuer: letsencrypt
```
paulopontesm: This one depends on us having https://github.com/jetstack/cert-manager installed in the cluster.

This component tries to automaticallly generate and validate a certificate for you.  Given that we don't mention it anywhere else, I think its better for us to leave this line out for now

codacy/values-production.yaml
```diff
@@ -66,22 +66,22 @@ codacy-ingress:
...
-      external-dns.alpha.kubernetes.io/hostname: '<--- codacy-app.dns.internal --->, <--- codacy-api.dns.internal --->' # Codacy application and API DNS 
+      external-dns.alpha.kubernetes.io/hostname: '<--- codacy-app.dns.internal --->, <--- codacy-api.dns.internal --->' # Codacy application and API hostnames
       external-dns.alpha.kubernetes.io/scope: public
       nginx.ingress.kubernetes.io/proxy-body-size: 128m
       nginx.ingress.kubernetes.io/ssl-redirect: "true"
       nginx.ingress.kubernetes.io/use-regex: "true"
     tls:
```
paulopontesm: We need to remove this if we dont put the certmanager

docs/install.md
```diff
@@ -2,19 +2,19 @@
 
 Follow the steps below to install Codacy on an existing Kubernetes cluster using the provided cloud native Helm chart.
 
-1.  Create a Kubernetes namespace called `codacy` that will group all cluster resources related to Codacy.
```
paulopontesm: Our auto formaters work differently. Are you using remark? If so, we need to put a config file on the repo to make sure we format the same when we save

codacy/values-production.yaml
joaocosta-codacy: We will not be using https (tls/ssl) on our self hosted installations. I guess this should also be dropped out for now? 
(this comment applies to all other values files as well :) )

codacy/values-production.yaml
paulopontesm: If we remove this we need to add to the docs how to get the IP/address of the ingress for people to setup their DNS.
@joaocosta-codacy can you help build a command that gets this value from the ingress?

docs/install.md
pauloribeiro-codacy: @paulopontesm and @joaocosta-codacy, could you please describe briefly the setup that you're using?

I'm used to having [markdownlint](https://github.com/DavidAnson/markdownlint) as a Visual Studio Code extension to validate Markdown during the authoring of the files. This linter seems to be a bit more prescriptive on enforcing a certain Markdown style than remark-lint and it plays well with MkDocs (it's even the [linter used in the MkDocs project](https://github.com/mkdocs/mkdocs/blob/master/.markdownlintrc) itself).

However, since Codacy includes the remark-lint as the Markdown linter, I'm thinking that I should be using that one instead as well. So, let me know how you're currently using this linter so that I can adapt on my side as well as create a configuration file to enforce a few style options.

pauloribeiro-codacy: I added a .remarkrc configuration file so that our Markdown style becomes consistent (https://github.com/codacy/chart/pull/293/commits/7963d611089f595d5853da112a1e5651c2837888). Hopefully this won't introduce too much noise in the diffs between files, if so, it may be better to remove the configuration file for now and only add it at a later time.

paulopontesm:
Can you add some final step on installation document on how to get to the UI if I have the ingress properly configured.

Maybe some kind of `kubectl get ingress .... `

codacy/values.yaml
```diff
@@ -263,14 +263,10 @@ codacy-ingress:
   create: true
   ingress:
     hostname:
-      app: <codacy-app.dns.internal>
-      api: <codacy-api.dns.internal>
+      app: "codacy.example.com"
```
joaocosta-codacy: if this line and the one below change, then you must update the .doks/Makefile `update_ingress_values` `sed` commands to match the changed syntax

paulopontesm: we should never ever edit this document. I will remove the line that edits this document

PR #295, opened 2020-03-12
crow enable /monitoring path by default

joaocosta-codacy:


codacy/values-production.yaml
```diff
@@ -240,6 +240,10 @@ crow:
     requests:
       cpu: 0.5
       memory: 1Gi
+  config:
```
paulopontesm: this is already the default because you put it in the `values.yaml` dosent need to be here

PR #297, opened 2020-03-12
Enable crow in chart through codacy-ingress

joaocosta-codacy:


docs/configuration/monitoring.md
```diff
@@ -58,6 +58,28 @@ Crow is a visualization tool that displays information about the projects and jo
 
 The Crow tool is installed alongside Codacy after the helm chart is deployed to the cluster.
 
+By default, the `crow` tool can be accessed through the `/monitoring` path of the url pointing to your Codacy installation, e.g. `http://codacy.company.org/monitoring`.
+It is possible for you to configure this path through the `values.yaml` file in the `crow` and `codacy-ingress` sections. You must also let `crow` know which is your codacy's url, so that it can properly anchor links to your projects.
+
+```yaml
+  crow:
+  config:
```
paulopontesm: I would remove this and just add the news paths on the table in the main README.md. Only keep here what is essential

docs/configuration/monitoring.md
```diff
@@ -58,6 +58,28 @@ Crow is a visualization tool that displays information about the projects and jo
...
+    play:
+      http:
+        context: <--- crow path --->
+    codacy:
+      url: <--- codacy url --->
+    crow:
+      url: <--- crow url --->
+
+  codacy-ingress:
```
paulopontesm: same as above

codacy/values-production.yaml
```diff
@@ -66,23 +66,28 @@ codacy-ingress:
...
       nginx.ingress.kubernetes.io/ssl-redirect: "true"
       nginx.ingress.kubernetes.io/use-regex: "true"
     tls:
       hosts:
-        - host: <--- codacy-app.dns.internal ---> # Codacy application DNS 
-        - host: <--- codacy-api.dns.internal ---> # Codacy API DNS 
+        - host: <--- codacy-app.dns.internal ---> # Codacy application DNS
+        - host: <--- codacy-api.dns.internal ---> # Codacy API DNS
       secretName: codacy-ingress-tls-secret
+    extraPaths:
```
paulopontesm: this should be just the default in the values.yaml.. we dont need to override it here

codacy/values-production.yaml
```diff
@@ -66,23 +66,28 @@ codacy-ingress:
   create: true
   ingress:
     hostname:
-      app: <--- codacy-app.dns.internal ---> # Codacy application DNS 
+      app: <--- codacy-app.dns.internal ---> # Codacy application DNS
       api: <--- codacy-api.dns.internal ---> # Codacy API DNS
     annotations:
       kubernetes.io/ingress.class: nginx
       kubernetes.io/tls-acme: "true"
       certmanager.k8s.io/acme-challenge-type: http01
```
paulopontesm: If you are changing this lets remove any certmanager and external dns reference from this ingress. 

Also any "tls thing"

.doks/doks-cluster/Makefile
```diff
@@ -34,27 +34,34 @@ setup_vars:
 	echo "num_nodes = $(NUM_NODES)" >> $(VARS_FILE)
 	echo "cluster_name = \"$(DOKS_CLUSTER_NAME)\"">> $(VARS_FILE)
 
+.PHONY: start_cluster
+start_cluster: provision_cluster set_cluster_context helm_init_upgrade install_nginx
+
+.PHONY: provision_cluster
+provision_cluster: setup_vars create_terraformrc
+# echo '1' | $(TF) init -input=false -reconfigure
```
paulopontesm: can you write line before this saying why this is commented plz?
Like, "Currently the tf state is not in sync with the infrastructure, we need to tweak it before we can have CD for this"

codacy/values.yaml
```diff
@@ -260,26 +260,25 @@ engine:
...
-    #   external-dns.alpha.kubernetes.io/scope: public
-    #   nginx.ingress.kubernetes.io/proxy-body-size: 128m
-    #   nginx.ingress.kubernetes.io/ssl-redirect: "true"
-    #   nginx.ingress.kubernetes.io/use-regex: "true"
-    # tls:
-    #   hosts:
-    #     - host: <codacy-app.dns.internal>
-    #     - host: <codacy-ap1.dns.internal>
-    #   secretName: codacy-ingress-tls-secret
+    hostname:
```
paulopontesm: I don't think we should uncomment the hostname, annotations and tls part on the values.yaml. They are here for reference.

docs/configuration/monitoring.md
```diff
@@ -58,6 +58,11 @@ Crow is a visualization tool that displays information about the projects and jo
 
 The Crow tool is installed alongside Codacy after the helm chart is deployed to the cluster.
 
+The `crow` tool can be accessed through the `/monitoring` path of the url pointing to your Codacy installation, e.g. `http://codacy.company.org/monitoring`.
```
rtfpessoa: Should't we use the global codacy url instead of `crow.config.codacy.url`? Also it would be cool to pass the `crow.config.crow.url` to a global one for the website to have the monitoring link back.

codacy/values.yaml
```diff
@@ -260,26 +260,25 @@ engine:
...
+      external-dns.alpha.kubernetes.io/hostname: <codacy-app.dns.internal>, <codacy-api.dns.internal>
+      external-dns.alpha.kubernetes.io/scope: public
+      nginx.ingress.kubernetes.io/proxy-body-size: 128m
+      nginx.ingress.kubernetes.io/ssl-redirect: "true"
+      nginx.ingress.kubernetes.io/use-regex: "true"
+    extraPaths:
+      - backend:
+          serviceName: codacy-crow
+          servicePort: http
+        path: /monitoring
```
rtfpessoa: Can we make this an internal variable since it needs to match?

docs/configuration/monitoring.md
paulopontesm: yep.. agree.. that would simplify the solution.
You need to change the base chart tough @joaocosta-codacy

joaocosta-codacy: good call. 
I am changing this in crow so that the crow.config.codacy.url points to .Values.global.codacy.url in the configmap template

codacy/values-production.yaml
```diff
@@ -66,23 +66,15 @@ codacy-ingress:
   create: true
   ingress:
     hostname:
-      app: <--- codacy-app.dns.internal ---> # Codacy application DNS
```
paulopontesm: it would be nice if this matched the example from `global.codacy.url`

.doks/Makefile
```diff
@@ -24,12 +24,16 @@ define call_helm_install
...
 	--set global.codacy.url=${4} \
+	--set global.codacy.crow.url=${4}/monitoring \
 	--set global.codacy.backendUrl=${4} \
 	--set listener.nfsserverprovisioner.storageClass.name="${1}-listener-cache-class" \
 	--set global.features.cloneSubmodules=true \
 	--set global.bitbucketEnterprise.consumerPublicKey=$${BITBUCKET_ENTERPRISE_PUBLIC_KEY} \
 	--set global.bitbucketEnterprise.consumerPrivateKey=$${BITBUCKET_ENTERPRISE_PRIVATE_KEY} \
-	--set worker-manager.config.codacy.worker.image=${5}
+	--set worker-manager.config.codacy.worker.image=${5} \
+	--set crow.config.crow.url=${4}/monitoring \
```
rtfpessoa: Why do we need two crow urls?

docs/configuration/monitoring.md
```diff
@@ -58,6 +58,17 @@ Crow is a visualization tool that displays information about the projects and jo
...
+You must set the `crow.config.codacy.url` and `crow.config.crow.url` values in your `values.yaml` file so that anchor links to your projects can be properly established inside `crow`.
+
+You must also provide a `CROW_PASSWORD` environment variable for your Crow password:
```
paulopontesm: I don't think this part makes sense here. This is a good section to be in the readme of crow tought

.doks/Makefile
joaocosta-codacy: this has been fixed on codacy-website's side. 👍

docs/configuration/monitoring.md
joaocosta-codacy: But how are we going to explain on the self-hosted docs that the customer needs to configure that variable to have the credentials configured?

codacy/values.yaml
joaocosta-codacy: We can not have variables in values files for now. I would say we move forward with this for now and come back to this issue later. Maybe we can just use a macro + sed to replace the value on both places.

README.md
```diff
@@ -192,6 +192,7 @@ The following parameters are specific to each Codacy component.
 | `crow.replicaCount`                                 | Number of replicas                                                                                  | `1`                                          |
 | `crow.image.repository`                             | Image repository                                                                                    | from dependency                              |
 | `crow.image.tag`                                    | Image tag                                                                                           | from dependency                              |
+| `crow.config.crow.url`                              | Url pointing to crow                                                                                | `nil`                                        |
```
rtfpessoa: Shouldn't this be global?

joaocosta-codacy: Crow didn't have this url exposed through global, which is why this isn't global on this chart. Should we change crow to have this url configured from .Values.global.crow.url instead, and reproduce that here?

joaocosta-codacy: Thoughts @rtfpessoa @paulopontesm ?

rtfpessoa: yes, that is probably the best idea. Since it is used in website and crow.
I would say all urls should probably be in global.

joaocosta-codacy: Updated in crow 4.9.1 👍

README.md
```diff
@@ -59,6 +59,7 @@ Global parameters apply to all sub-charts and make it easier to configure resour
 | ----------------------------------------------- | ------------------------------------------------------------------------------------------------------------ | --------------- |
 | `global.codacy.url`                             | Hostname to your Codacy installation                                                                         | `nil`           |
 | `global.codacy.backendUrl`                      | Hostname to your Codacy installation                                                                         | `nil`           |
+| `global.codacy.crow.crow.url`                   | Hostname to Crow within your Codacy installation                                                             | `nil`           |
```
rtfpessoa: crow.crow?

PR #300, opened 2020-03-16
Change fluentd lifecycle configuration CY-1139

jllopes:


README.md
```diff
@@ -198,7 +198,7 @@ The following parameters refer to components that are not internal to Codacy, bu
 | Parameter                        | Description                                                                                            | Default    |
 | -------------------------------- | ------------------------------------------------------------------------------------------------------ | ---------- |
 | `fluentdoperator.enable`         | Enable fluentd operator. It gathers logs from Codacy so that you can send it to our support if needed. | `false`    |
```
afsmeira: Should the default be true?

jllopes: I agree with this, @paulopontesm @joaocosta-codacy what do you guys think?

joaocosta-codacy: I would say yes to having it enabled by default. We do not want to put customers in a position where "we could help you if you had the logs". 
Do we know how many logs are exported/how big the logfiles are? My main concern is just going above a pvc limit for having verboseness enabled by default and the cluster getting clogged/slow/broken. But i do not possess that kind of information (yet) :) @paulopontesm ?

codacy/templates/fluentd/fluentd-conf.yaml
```diff
@@ -43,10 +43,10 @@ data:
       s3_object_key_format %{path}%{index}.%{file_extension}
 
       # We can't use this due to a bug in minio
-      # <bucket_lifecycle_rule>
-      #   id test-policy
-      #   expiration_days {{ .Values.fluentdoperator.expirationDays }}
-      # </bucket_lifecycle_rule>
+      <bucket_lifecycle_rule>
```
paulopontesm: don't we need to have minio bumped to have this working? I tought the version we were using had some kid on bug

README.md
paulopontesm: Ok. 
We decided not to enable it by default because it was kind of intrusive since it creates a daemon set that attaches itself to the /var/log in all machines.
The idea was to have it disabled here and enabled in the `values-production.yaml`.

But I am confortable with having this enabled here. Just make sure you update the `values-production.yaml` so that it overrides values that are not default (remove the enable from there)

codacy/templates/fluentd/fluentd-conf.yaml
jllopes: I tested this by hand in our sandbox environment, and it seems that it is working as expected

README.md
```diff
@@ -197,8 +197,8 @@ The following parameters refer to components that are not internal to Codacy, bu
 
 | Parameter                        | Description                                                                                            | Default    |
 | -------------------------------- | ------------------------------------------------------------------------------------------------------ | ---------- |
-| `fluentdoperator.enable`         | Enable fluentd operator. It gathers logs from Codacy so that you can send it to our support if needed. | `false`    |
-| `fluentdoperator.expirationDays` | Number of days to retain logs. More time uses more disk on minio.                                      | `14`       |
+| `fluentdoperator.enable`         | Enable fluentd operator. It gathers logs from Codacy so that you can send it to our support if needed. | `true`     |
```
rtfpessoa: :+1:

codacy/templates/fluentd/fluentd-conf.yaml
```diff
@@ -43,10 +43,10 @@ data:
...
+        id test-policy
```
rtfpessoa: Can we have a proper name here. Like `retention-policy`

codacy/templates/fluentd/fluentd-conf.yaml
```diff
@@ -43,10 +43,10 @@ data:
       s3_object_key_format %{path}%{index}.%{file_extension}
 
       # We can't use this due to a bug in minio
```
afsmeira: This should have been removed 😅

rtfpessoa: Hahaha Nice catch

PR #306, opened 2020-03-18
doc: Add kubernetes cheatsheet

rtfpessoa:


docs/extra/k8s-cheatsheet.md
```diff
@@ -0,0 +1,102 @@
+# Kubernetes Cheatsheet
+
+## Session Manager SSH
+
+When using AWS Session Manager since the CLI is very limited you will benefit from using this alias:
```
afsmeira: this -> these

docs/extra/k8s-cheatsheet.md
```diff
@@ -0,0 +1,102 @@
...
+```
+
+### Check uninstall process
+ps aux | grep -i kubectl
+
+## Get logs of a service
+
+```bash
+kubectl get services
+kubectl logs svc/codacy-hotspots-api
```
afsmeira: I would replace `codacy-hotspots-api` by `<service-name>`.

docs/extra/k8s-cheatsheet.md
```diff
@@ -0,0 +1,102 @@
...
+```
+
+## Edit configmap
+
+```bash
+kubectl get configmaps
+kubectl edit configmap codacy-fluentd-config
```
afsmeira: Same here, and throughout this file.

docs/extra/k8s-cheatsheet.md
rtfpessoa:
```sugestion
When using AWS Session Manager since the CLI is very limited you will benefit from using these aliases:
```

rtfpessoa: Next time do code suggestion please. Check bellow for example.

docs/extra/k8s-cheatsheet.md
```diff
@@ -0,0 +1,102 @@
...
+```
+
+### Check uninstall process
+ps aux | grep -i kubectl
+
+## Get logs of a service
+
+```bash
+kubectl get services
+kubectl logs svc/codacy-hotspots-api
```
rtfpessoa:
```suggestion
kubectl logs svc/<service-name>
```

docs/extra/k8s-cheatsheet.md
```diff
@@ -0,0 +1,98 @@
...
+sleep 5
+kubectl -n codacy delete pod $(kubectl -n codacy get pod -o jsonpath='{.items[*].metadata.name}') --force --grace-period=0
+kubectl -n codacy get pod &
+kubectl -n codacy get pvc &
+kubectl -n codacy get pv  &
+kubectl -n codacy get job &
+```
+
+### Check uninstall process
+ps aux | grep -i kubectl
```
pauloribeiro-codacy: Include this in a bash codeblock?

docs/extra/k8s-cheatsheet.md
```diff
@@ -0,0 +1,98 @@
...
+kubectl -n codacy patch pv -p '{"metadata":{"finalizers":null}}' $(kubectl -n codacy get pv -o jsonpath='{.items[*].metadata.name}')
+sleep 5
+kubectl -n codacy delete pod $(kubectl -n codacy get pod -o jsonpath='{.items[*].metadata.name}') --force --grace-period=0
+kubectl -n codacy get pod &
+kubectl -n codacy get pvc &
+kubectl -n codacy get pv  &
+kubectl -n codacy get job &
+```
+
+### Check uninstall process
```
pauloribeiro-codacy:
```suggestion
### Check uninstall was successful
```

(I'm not sure if it's too important, but this check does not work for microk8s.)

docs/extra/k8s-cheatsheet.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Open shell inside container
+
+```bash
+kubectl exec -it daemonset/<daemonset-name> -c <container-name> sh
+
+# or
+
+kubectl exec -it deployment/<deployment-name> sh
+```
```
pauloribeiro-codacy: Separate these into two bash codeblocks? I think the "or" will be clearer / more visible.

docs/extra/k8s-cheatsheet.md
```diff
@@ -0,0 +1,98 @@
...
+
+### Update
+
+```bash
+(cd chart; sudo git fetch --all --prune --tags; sudo git reset --hard origin/<YOUR-BRANCH>;)
+helm dep build ./chart/codacy
+helm upgrade --install codacy ./chart/codacy/ --namespace codacy --atomic --timeout=300 --values ./<YOUR-VALUES-FILE>
+```
+
+## How to clean the namespace
```
pauloribeiro-codacy: For the sake of consistency:

```suggestion
## Clean the namespace
```

docs/extra/k8s-cheatsheet.md
```diff
@@ -0,0 +1,98 @@
+# Kubernetes Cheatsheet
+
+## Session Manager SSH
+
+When using AWS Session Manager since the CLI is very limited you will benefit from using these aliases:
+```bash
+alias kubectl='sudo microk8s.kubectl -n codacy'
```
pauloribeiro-codacy: This will collide with the existing (customer-facing) docs because we're including the "-n codacy" in the instructions.

It would be better like this:

```suggestion
alias kubectl='sudo microk8s.kubectl'
```

rtfpessoa: @afsmeira asked this before and in this specific case this is related to our microk8s.
Do you think I should make it explicit or move it to the other repo?

docs/extra/k8s-cheatsheet.md
rtfpessoa: Any special reason why? The reason you have a code block is usually to have multiple code blocks.

paulopontesm:
> I agree with @afsmeira that this is a great initiative. There are very useful commands in there!
> 
> I'm just not sure if this is meant for our internal use or for customers. If it is for customers I think we need to improve / check some commands because they may not apply to all kind of infrastructures (Kubernetes or MicroK8s, running on VMs or AWS or Azure...). There are lots of different combinations... :(

This is really good!!! I saw this section on the elastic documentation that looks a lot like what is in here and seemed really helpful

https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-troubleshooting.html

docs/extra/k8s-cheatsheet.md
pauloribeiro-codacy: The user will run either one command or the other, correct? So it will make it easier to copy + paste only the intended command.

rtfpessoa: These are just examples. You can copy each individually.
None of these should be run together.
These are all commands you need to tweak to do something.

In that case the idea is to show one where we have multiple containers and another when you don't.

docs/extra/k8s-cheatsheet.md
pauloribeiro-codacy: OK, if this is specific to MicroK8s, then I think the only command where there will be a collision because we're specifying the namespace twice would be on the second step here:

https://github.com/codacy/chart/blob/microk8s-install-docs/docs/install.md

Notice that the end of the MicroK8s quick start guide refers to these instructions:

https://github.com/codacy/chart/blob/microk8s-install-docs/docs/install.md

docs/extra/k8s-cheatsheet.md
```diff
@@ -0,0 +1,104 @@
...
+
+```bash
+kubectl exec -it daemonset/<daemonset-name> -c <container-name> sh
+
+# or
+
+kubectl exec -it deployment/<deployment-name> sh
+```
+
+## microk8s
```
rtfpessoa: @pauloribeiro-codacy @afsmeira I moved this to the bottom and made it explicit that it is for microk8s.
Let me know if you think this should go in other place

pauloribeiro-codacy: I think it's more explicit now, yes. :+1:

docs/extra/k8s-cheatsheet.md
rtfpessoa: I changed it to be configured like the rest of the commands. What do you think?

rtfpessoa:
> @afsmeira But I think that we should have a link to this page somewhere, otherwise it will get lost. Not sure where the link should exist, but I think it should be in the homepage or close to the home page (perharps in a page that is linked from the homepage).

Do you have a place to link that?
In general not being linked is not a problem since most of the times (at least for me) I navigate with the search.

afsmeira:
> Do you have a place to link that?
> In general not being linked is not a problem since most of the times (at least for me) I navigate with the search.

Screenshot 2020-03-19 at 11 01 52

Looking at the TOC, I would suggest a page in the 'Extra' section or a page in the top-level.

rtfpessoa:
@afsmeira but that is where it will appear automatically.
My understanding is that TOC is generated from the folder structure. It is not manual.
Am I wrong @paulopontesm ?

rtfpessoa:
proof
Yup, just tested locally. @afsmeira

docs/extra/k8s-cheatsheet.md
pauloribeiro-codacy: Exactly... The MkDocs template we're using includes a button that allows copying the content of codeblocks for convenience (you can see this working [here](https://codacy.github.io/chart/install/)):

image

What I was trying to say is that if we have separate codeblocks for each command that the user must choose from, it becomes more user friendly.

rtfpessoa: ok, that sounds like a good reason

PR #307, opened 2020-03-18
docs: Add log extraction script [CY-1140]

afsmeira:
To improve the life of our K8s clients, a log extraction script was created.
The script also compacts the logs using `zip`.

Ideally this script will be short live since one of two things happen:
* A UI is created to log extraction
* Selfhosted K8s and MicroK8s will include a data pipeline platform (such as
ELK), where logs will be available.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,21 @@
+# This file is for unifying the coding style for different editors and IDEs
+# editorconfig.org
+
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.{md,markdown,mdown,mkdn,mkd,mdwn,mkdown,ron}]
+indent_size = 4
+
+[Makefile]
+indent_style = tab
+
+[*.{sh,bash}]
+indent_style = tab
--- a/docs/configuration/extract-codacy-logs.sh
+++ b/docs/configuration/extract-codacy-logs.sh
@@ -24,6 +24,7 @@ date_days_ago() {
 
 cleanup()
 {
+    echo "Log files compression failed, exiting..."
     if [ -d "$LOGS_DIR" ]; then
         rm -r $LOGS_DIR &>/dev/null
     fi
@@ -62,7 +63,6 @@ echo "Checking if zip is installed..."
 if ! zip --version &>/dev/null; then
     echo "zip not installed"
     echo "Please install zip"
-    echo "Log files extraction failed, exiting..."
     exit 4
 fi
 
@@ -72,9 +72,8 @@ echo "Checking if kubectl is installed..."
 KUBECTL=$(which kubectl || which microk8s.kubectl)
 if [ $? -ne 0 ]; then
     echo "kubectl not installed"
-    echo "Please install kubectl version specified in Codacy's documentation (see here - https://codacy.github.io/chart/#preparing-to-install-codacy) or the version used when installing your cluster"
+    echo "Please install kubectl version specified in Codacy's documentation (see here - https://codacy.github.io/chart/install/) or the version used when installing your cluster"
     echo "To install kubectl see https://kubernetes.io/docs/tasks/tools/install-kubectl/ (or https://microk8s.io/docs/ if you are running a microk8s kubernetes cluster)"
-    echo "Log files extraction failed, exiting..."
     exit 4
 fi
 
@@ -83,21 +82,18 @@ echo "Checking access to kubernetes cluster..."
 KUBE_CTX=$($KUBECTL config current-context)
 if [ $? -ne 0 ]; then
     echo "No kubernetes cluster context configured"
-    echo "Log files extraction failed, exiting..."
     exit 5
 fi
 
 read -p "Is '$KUBE_CTX' the correct kubernetes cluster for log extraction? ([yes]/no)" ANSWER
 if [[ ! "$ANSWER" =~ ^y(es)?$ ]]; then
     echo "Please configure correctly your current kubernetes cluster"
-    echo "Log files extraction failed, exiting..."
     exit 6
 fi
 
 # Create temporary directory for copying logs
 if ! mkdir -p $LOGS_DIR; then
     echo "Failed to create temporary directory $LOGS_DIR , for log files extraction"
-    echo "Log files extraction failed, exiting..."
     exit 7
 fi
 
@@ -107,7 +103,6 @@ LOGS_POD_NAME=$($KUBECTL get pods -n $NAMESPACE -l app=minio -o jsonpath='{.item
 if [ $? -ne 0 ]; then
     echo "Failed to get the name of the logs kubernetes pod, for namespace $NAMESPACE"
     echo "Are you sure you are in the right kubernetes cluster context?"
-    echo "Log files extraction failed, exiting..."
     exit 8
 fi
 
@@ -116,7 +111,6 @@ echo "Extracting log files..."
 if ! $KUBECTL cp $NAMESPACE/$LOGS_POD_NAME:/export/logs $LOGS_DIR; then
     echo "Failed to extract log files from kubernetes pod $LOGS_POD_NAME to local directory $LOGS_DIR"
     echo "Are you sure you are in the right kubernetes cluster context?"
-    echo "Log files extraction failed, exiting..."
     exit 9
 fi
 
@@ -135,7 +129,6 @@ if [ -n "$DAYS" ]; then
         if [ $? -ne 0 ]; then
             echo "Failed to compress logs (located in $LOGS_DIR) to a ZIP file"
             echo "If this step continues to fail, you can compress the files manually"
-            echo "Log files compression failed, exiting..."
             exit 10
         fi
 
@@ -147,7 +140,6 @@ else
     if ! zip -r9 codacy_logs_$CURRENT_DATE_TIME.zip $LOGS_DIR; then
         echo "Failed to compress logs (located in $LOGS_DIR) to a ZIP file"
         echo "If this step continues to fail, you can compress the files manually"
-        echo "Log files compression failed, exiting..."
         exit 10
     fi
 fi
--- a/docs/configuration/extract-codacy-logs.sh
+++ b/docs/configuration/extract-codacy-logs.sh
@@ -1,120 +1,100 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 #### Helper functions ####
 
 usage()
 {
-  echo "Usage: $0 -n NAMESPACE "
-  exit 2
-}
-
-set_variable()
-{
-  local varname=$1
-  shift
-  if [ -z "${!varname}" ]; then
-    eval "$varname=\"$@\""
-  else
-    echo "Error: $varname already set"
-    usage
-  fi
+    echo "Usage: $0 -n <NAMESPACE> "
+    exit 3
 }
 
 cleanup()
 {
-	if [ -z $LOGS_DIR ]; then
-		rm -r $LOGS_DIR &>/dev/null
-	fi
+    if [ -d "$LOGS_DIR" ]; then
+        rm -r $LOGS_DIR &>/dev/null
+    fi
 }
-
+trap cleanup EXIT
 
 #### Log extraction script ####
 
 while getopts "n:" option; do
-	case ${option} in
-		n )
-			set_variable NAMESPACE ${OPTARG}
-			;;
-		\? )
-			usage
-			;;
-	esac
+    case ${option} in
+        n )
+            NAMESPACE=${OPTARG}
+            ;;
+        \? )
+            usage
+            ;;
+    esac
 done
 
 [ -z "$NAMESPACE" ] && usage
 
 CURRENT_DATE_TIME=$(date "+%Y%m%d-%H%M%S")
-LOGS_DIR=/tmp/codacy_logs_$CURRENT_DATE_TIME
+LOGS_DIR="$(mktemp -d)/codacy_logs_$CURRENT_DATE_TIME"
 
 echo "Starting log files extraction"
 
 # Check if kubectl is available
 echo "Checking if kubectl is installed..."
-kubectl version &>/dev/null
-if [ $? -ne 0 ]; then
-	echo "kubectl not installed"
-	echo "Please install kubectl version specified in Codacy's documentation (see here - https://codacy.github.io/chart/#preparing-to-install-codacy) or the version used when installing your cluster"
-	echo "To install kubectl see https://kubernetes.io/docs/tasks/tools/install-kubectl/"
-	echo "Log files extraction failed, exiting..."
-	exit $?
+if ! kubectl version &>/dev/null; then
+    echo "kubectl not installed"
+    echo "Please install kubectl version specified in Codacy's documentation (see here - https://codacy.github.io/chart/#preparing-to-install-codacy) or the version used when installing your cluster"
+    echo "To install kubectl see https://kubernetes.io/docs/tasks/tools/install-kubectl/"
+    echo "Log files extraction failed, exiting..."
+    exit 4
 fi
 
 # Check current cluster context
 echo "Checking access to kubernetes cluster..."
 KUBE_CTX=$(kubectl config current-context)
 if [ $? -ne 0 ]; then
-	echo "No kubernetes cluster context configured"
-	echo "Log files extraction failed, exiting..."
-	exit $?
+    echo "No kubernetes cluster context configured"
+    echo "Log files extraction failed, exiting..."
+    exit 5
 fi
 
-echo "Is '$KUBE_CTX' the correct kubernetes cluster for log extraction? (yes/no)"
-read ANSWER
-if [ "$ANSWER" != "yes" ]; then
-	echo "Please configure correctly your current kubernetes cluster"
-	echo "Log files extraction failed, exiting..."
-	exit 2
+read -p "Is '$KUBE_CTX' the correct kubernetes cluster for log extraction? ([yes]/no)" ANSWER
+if [[ ! "$ANSWER" =~ ^y(es)?$ ]]; then
+    echo "Please configure correctly your current kubernetes cluster"
+    echo "Log files extraction failed, exiting..."
+    exit 6
 fi
 
 # Create temporary directory for copying logs
-mkdir $LOGS_DIR
-if [ $? -ne 0 ]; then
-	echo "Failed to create temporary directory $LOGS_DIR , for log files extraction"
-	echo "Log files extraction failed, exiting..."
-	exit $?
+if ! mkdir -p $LOGS_DIR; then
+    echo "Failed to create temporary directory $LOGS_DIR , for log files extraction"
+    echo "Log files extraction failed, exiting..."
+    exit 7
 fi
 
 # Get pod name for the logs pod
 echo "Checking if the logs kubernetes pod exists..."
 LOGS_POD_NAME=$(kubectl get pods -n $NAMESPACE -l app=minio -o jsonpath='{.items[*].metadata.name}')
 if [ $? -ne 0 ]; then
-	echo "Failed to get the name of the logs kubernetes pod, for namespace $NAMESPACE"
-	echo "Are you sure you are in the right kubernetes cluster context?"
-	echo "Log files extraction failed, exiting..."
-	cleanup
-	exit $?
+    echo "Failed to get the name of the logs kubernetes pod, for namespace $NAMESPACE"
+    echo "Are you sure you are in the right kubernetes cluster context?"
+    echo "Log files extraction failed, exiting..."
+    exit 8
 fi
 
 # Copy logs to local filesystem
 echo "Extracting log files..."
-kubectl cp $NAMESPACE/$LOGS_POD_NAME:/export/logs $LOGS_DIR
-if [ $? -ne 0 ]; then
-	echo "Failed to extract log files from kubernetes pod $LOGS_POD_NAME to local directory $LOGS_DIR"
-	echo "Are you sure you are in the right kubernetes cluster context?"
-	echo "Log files extraction failed, exiting..."
-	cleanup
-	exit $?
+if ! kubectl cp $NAMESPACE/$LOGS_POD_NAME:/export/logs $LOGS_DIR; then
+    echo "Failed to extract log files from kubernetes pod $LOGS_POD_NAME to local directory $LOGS_DIR"
+    echo "Are you sure you are in the right kubernetes cluster context?"
+    echo "Log files extraction failed, exiting..."
+    exit 9
 fi
 
 # Compress logs in ZIP file (-9 is maximum, slowest, compression)
 echo "Compressing extracted log files..."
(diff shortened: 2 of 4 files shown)


docs/configuration/extract-codacy-logs.sh
```diff
@@ -0,0 +1,112 @@
...
+if [ $ANSWER != "yes" ]; then
+	echo "Please configure correctly your current kubernetes cluster"
+	echo "Log files extraction failed, exiting..."
+	exit 2
+fi
+
+# Create temporary directory for copying logs
+mkdir $LOGS_DIR
+if [ $? -ne 0 ]; then
+	echo "Failed to create temporary directory $LOGS_DIR , for log files extraction"
```
afsmeira: If this happens we should probably continue in the local folder.

docs/configuration/extract-codacy-logs.sh
```diff
@@ -0,0 +1,112 @@
...
+    usage
+  fi
+}
+
+
+#### Log extraction script ####
+
+while getopts "n:" option; do
+	case ${option} in
+		n )
```
afsmeira: We should add an optional flag for only the last 3 days of logs, because in most cases we will not need the full 7 days.

docs/configuration/extract-codacy-logs.sh
```diff
@@ -0,0 +1,112 @@
...
+	echo "Log files extraction failed, exiting..."
+	exit $?
+fi
+
+# Copy logs to local filesystem
+echo "Extracting log files..."
+kubectl cp $NAMESPACE/$LOGS_POD_NAME:/export/logs $LOGS_DIR
+if [ $? -ne 0 ]; then
+	echo "Failed to extract log files from kubernetes pod $LOGS_POD_NAME to local directory $LOGS_DIR"
+	echo "Are you sure you are in the right cluster context?"
```
afsmeira: For consistency, this should be "(...) the right kubernetes cluster context?"

docs/configuration/extract-codacy-logs.sh
```diff
@@ -0,0 +1,120 @@
+#!/bin/bash
```
rtfpessoa: Please use env to get the bash instead of targeting a specific binary.
```suggestion
#!/usr/bin/env bash
```

docs/configuration/extract-codacy-logs.sh
```diff
@@ -0,0 +1,120 @@
...
+
+#### Helper functions ####
+
+usage()
+{
+  echo "Usage: $0 -n NAMESPACE "
```
rtfpessoa:
```suggestion
  echo "Usage: $0 -n <NAMESPACE> "
```

docs/configuration/extract-codacy-logs.sh
```diff
@@ -0,0 +1,120 @@
...
+    eval "$varname=\"$@\""
+  else
+    echo "Error: $varname already set"
+    usage
+  fi
+}
+
+cleanup()
+{
+	if [ -z $LOGS_DIR ]; then
```
rtfpessoa: Why is the indentation here different? 
Can you add an `.editorconfig` yo the root of this project?

docs/configuration/extract-codacy-logs.sh
```diff
@@ -0,0 +1,120 @@
...
+  echo "Usage: $0 -n NAMESPACE "
+  exit 2
+}
+
+set_variable()
+{
+  local varname=$1
+  shift
+  if [ -z "${!varname}" ]; then
+    eval "$varname=\"$@\""
```
rtfpessoa: Why do you need eval?

rtfpessoa: Why $@ instead of $1?

docs/configuration/extract-codacy-logs.sh
```diff
@@ -0,0 +1,120 @@
+#!/bin/bash
+
+#### Helper functions ####
+
+usage()
+{
+  echo "Usage: $0 -n NAMESPACE "
+  exit 2
```
rtfpessoa: Should we use 2 here? Check http://www.tldp.org/LDP/abs/html/exitcodes.html

docs/configuration/extract-codacy-logs.sh
rtfpessoa: `-z` check if the variable is empty. Seems like this is not what you want here.
Shouldn't it be something like:
```suggestion
	if [ -d "$LOGS_DIR" ]; then
```

docs/configuration/extract-codacy-logs.sh
```diff
@@ -0,0 +1,120 @@
...
+		\? )
+			usage
+			;;
+	esac
+done
+
+[ -z "$NAMESPACE" ] && usage
+
+CURRENT_DATE_TIME=$(date "+%Y%m%d-%H%M%S")
+LOGS_DIR=/tmp/codacy_logs_$CURRENT_DATE_TIME
```
rtfpessoa:
```suggestion
LOGS_DIR="$(mktemp -d)/codacy_logs_$CURRENT_DATE_TIME"
```

docs/configuration/extract-codacy-logs.sh
```diff
@@ -0,0 +1,120 @@
...
+
+echo "Starting log files extraction"
+
+# Check if kubectl is available
+echo "Checking if kubectl is installed..."
+kubectl version &>/dev/null
+if [ $? -ne 0 ]; then
```
rtfpessoa: You can check the outcome directly like this, maybe yours is more explicit.
```suggestion
if ! kubectl version &>/dev/null; then
```

docs/configuration/extract-codacy-logs.sh
```diff
@@ -0,0 +1,120 @@
...
+echo "Checking access to kubernetes cluster..."
+KUBE_CTX=$(kubectl config current-context)
+if [ $? -ne 0 ]; then
+	echo "No kubernetes cluster context configured"
+	echo "Log files extraction failed, exiting..."
+	exit $?
+fi
+
+echo "Is '$KUBE_CTX' the correct kubernetes cluster for log extraction? (yes/no)"
+read ANSWER
```
rtfpessoa:
```suggestion
read -p "Is '$KUBE_CTX' the correct kubernetes cluster for log extraction? ([yes] / no)" ANSWER
```

docs/configuration/extract-codacy-logs.sh
```diff
@@ -0,0 +1,120 @@
...
+if [ "$ANSWER" != "yes" ]; then
```
rtfpessoa:
```suggestion
if [[ "$ANSWER" =~ ^y(es)?$ ]]; then
```

docs/configuration/extract-codacy-logs.sh
```diff
@@ -0,0 +1,120 @@
...
+	echo "Please configure correctly your current kubernetes cluster"
+	echo "Log files extraction failed, exiting..."
+	exit 2
```
rtfpessoa: Would be cool if we could get different exit codes for each error.

docs/configuration/extract-codacy-logs.sh
```diff
@@ -0,0 +1,120 @@
...
+fi
+
+# Create temporary directory for copying logs
+mkdir $LOGS_DIR
```
rtfpessoa:
```suggestion
mkdir -p $LOGS_DIR
```

docs/configuration/extract-codacy-logs.sh
```diff
@@ -0,0 +1,120 @@
...
+if [ $? -ne 0 ]; then
+	echo "Failed to create temporary directory $LOGS_DIR , for log files extraction"
+	echo "Log files extraction failed, exiting..."
+	exit $?
```
rtfpessoa: Here you already lost the code since you run other commands like echo

docs/configuration/extract-codacy-logs.sh
```diff
@@ -0,0 +1,120 @@
...
+zip -r9 codacy_logs_$CURRENT_DATE_TIME.zip $LOGS_DIR
+if [ $? -ne 0 ]; then
+	echo "Failed to compress logs (located in $LOGS_DIR) to a ZIP file"
+	echo "If this step continues to fail, you can compress the files manually"
+	echo "Log files compression failed, exiting..."	
+	exit $?
+fi
+
+echo "Log file extraction completed"
+cleanup
```
rtfpessoa: Can we put this in a trap like http://redsymbol.net/articles/bash-exit-traps/

docs/configuration/logging.md
```diff
@@ -35,8 +35,10 @@ To collect logs from your Codacy installation using fluentd change `fluentdopera
 
 The fluentd daemonset will send the logs to minio which is also installed by this chart.
 
-In order to send them to our support in case of problems, run the following command locally (replacing the `<namespace>` with the namespace in which Codacy was installed):
+To extract the logs and send them to Codacy's support team in case of problems, you can run the following command locally (replacing the `<namespace>` with the namespace in which Codacy was installed):
 
 ```bash
-kubectl cp <namespace>/$(kubectl get pods -n <namespace> -l app=minio -o jsonpath='{.items[*].metadata.name}'):/export/logs ./logs
+curl https://codacy.github.io/chart/configuration/extract-codacy-logs.sh > extract-codacy-logs.sh && chmod +x extract-codacy-logs.sh && extract-codacy-logs.sh -n <namespace>
```
rtfpessoa:
```suggestion
curl -fsSL https://codacy.github.io/chart/configuration/extract-codacy-logs.sh | bash -s -- -n <namespace>
```

rtfpessoa: We could use a shorter version here. Since the other is a single command this seems more concise.

docs/configuration/extract-codacy-logs.sh
rtfpessoa: hmm, I would say if this happens the machine is fucked up.
What would be another reason for a failure here?

docs/configuration/extract-codacy-logs.sh
```diff
@@ -0,0 +1,115 @@
...
+[ -z "$NAMESPACE" ] && usage
+
+CURRENT_DATE_TIME=$(date "+%Y%m%d-%H%M%S")
+LOGS_DIR="$(mktemp -d)/codacy_logs_$CURRENT_DATE_TIME"
+
+echo "Starting log files extraction"
+
+# Check if kubectl is available
+echo "Checking if kubectl is installed..."
+if ! kubectl version &>/dev/null; then
```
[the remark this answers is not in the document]
afsmeira: In that case I need the value returned from the command... I guess I could use `$?` if no other commands are issued in between.

docs/configuration/extract-codacy-logs.sh
```diff
@@ -0,0 +1,115 @@
...
+if [ $? -ne 0 ]; then
+    echo "Failed to get the name of the logs kubernetes pod, for namespace $NAMESPACE"
+    echo "Are you sure you are in the right kubernetes cluster context?"
+    echo "Log files extraction failed, exiting..."
+    exit 8
+fi
+
+# Copy logs to local filesystem
+echo "Extracting log files..."
+kubectl cp $NAMESPACE/$LOGS_POD_NAME:/export/fluentd-bucket $LOGS_DIR
```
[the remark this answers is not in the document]
afsmeira: He did. I messed up and commited this after changing it for testing.

docs/configuration/extract-codacy-logs.sh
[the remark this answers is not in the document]
afsmeira: I can't because `$?` has the exit code and not the command output. Maybe I could do it in a smarter way, but the way it is right now is simple enough and works.

docs/configuration/extract-codacy-logs.sh
afsmeira: Maybe some permissions issue? Not sure when you don't have permissions to write to `/tmp` so let's not try to be too smart about this. I'll mark it as resolved.

docs/configuration/extract-codacy-logs.sh
afsmeira: Anyway, in the beginning I had the expectation of using this more, but since it's only used once, I'll simplify.

docs/configuration/extract-codacy-logs.sh
```diff
@@ -0,0 +1,100 @@
...
+echo "Checking access to kubernetes cluster..."
+KUBE_CTX=$(kubectl config current-context)
+if [ $? -ne 0 ]; then
+    echo "No kubernetes cluster context configured"
+    echo "Log files extraction failed, exiting..."
+    exit 5
+fi
+
+read -p "Is '$KUBE_CTX' the correct kubernetes cluster for log extraction? ([yes]/no)" ANSWER
+if [[ "$ANSWER" =~ ^y(es)?$ ]]; then
```
rtfpessoa:
```suggestion
if [[ ! "$ANSWER" =~ ^y(es)?$ ]]; then
```

.editorconfig
```diff
@@ -0,0 +1,21 @@
+# This file is for unifying the coding style for different editors and IDEs
```
paulopontesm: I am using remark lint (the tool that codacy uses for linting) to format the markdown.

@pauloribeiro-codacy is putting a configuration for this in this other branch https://github.com/codacy/chart/pull/293/files

docs/configuration/extract-codacy-logs.sh
```diff
@@ -0,0 +1,100 @@
...
+if ! kubectl cp $NAMESPACE/$LOGS_POD_NAME:/export/logs $LOGS_DIR; then
+    echo "Failed to extract log files from kubernetes pod $LOGS_POD_NAME to local directory $LOGS_DIR"
+    echo "Are you sure you are in the right kubernetes cluster context?"
+    echo "Log files extraction failed, exiting..."
+    exit 9
+fi
+
+# Compress logs in ZIP file (-9 is maximum, slowest, compression)
+echo "Compressing extracted log files..."
+if ! zip -r9 codacy_logs_$CURRENT_DATE_TIME.zip $LOGS_DIR; then
```
paulopontesm: zip is not installed by default in ubuntu. Should we check for this?

docs/configuration/logging.md
```diff
@@ -35,8 +35,10 @@ To collect logs from your Codacy installation using fluentd change `fluentdopera
 
 The fluentd daemonset will send the logs to minio which is also installed by this chart.
 
-In order to send them to our support in case of problems, run the following command locally (replacing the `<namespace>` with the namespace in which Codacy was installed):
+To extract the logs and send them to Codacy's support team in case of problems, you can run the following command locally (replacing the `<namespace>` with the namespace in which Codacy was installed):
 
 ```bash
-kubectl cp <namespace>/$(kubectl get pods -n <namespace> -l app=minio -o jsonpath='{.items[*].metadata.name}'):/export/logs ./logs
+bash <(curl -fsSL https://codacy.github.io/chart/configuration/extract-codacy-logs.sh) -n <namespace>
```
paulopontesm: https://codacy.github.io/chart/configuration/extract-codacy-logs.sh  -> I don't think this url will work. Only the things inside the `docs/` directory stay available on the static page.

Check this branch: https://github.com/codacy/chart/tree/gh-pages

Also, check the coverage reporter. It also has a script in the repo that is being provided through github pages https://github.com/codacy/codacy-coverage-reporter/

.editorconfig
rtfpessoa: This is a universal file that most editors pick if there is no other tool

docs/configuration/logging.md
```diff
@@ -35,8 +35,10 @@ To collect logs from your Codacy installation using fluentd change `fluentdopera
 
 The fluentd daemonset will send the logs to minio which is also installed by this chart.
 
-In order to send them to our support in case of problems, run the following command locally (replacing the `<namespace>` with the namespace in which Codacy was installed):
+To extract the logs and send them to Codacy's support team in case of problems, you can run the following command locally (replacing the `<namespace>` with the namespace in which Codacy was installed):
```
paulopontesm: Can you put a note here saying that you should create an alias for microk8s? 
Like this: https://github.com/codacy/chart/blob/master/docs/infrastructure/microk8s-quickstart.md#5-installing-codacy

docs/configuration/extract-codacy-logs.sh
```diff
@@ -0,0 +1,133 @@
...
+if ! kubectl cp $NAMESPACE/$LOGS_POD_NAME:/export/fluentd-bucket $LOGS_DIR; then
+    echo "Failed to extract log files from kubernetes pod $LOGS_POD_NAME to local directory $LOGS_DIR"
+    echo "Are you sure you are in the right kubernetes cluster context?"
+    echo "Log files extraction failed, exiting..."
+    exit 9
+fi
+
+
+echo "Compressing extracted log files..."
+if [ ! -z $DAYS ]; then
```
rtfpessoa: When using variables is always good to double quote them to avoid expanding through spaces and having a weird behaviour. Also e opposite of `-z` is `-n`
```suggestion
if [ -n "$DAYS" ]; then
```

docs/configuration/extract-codacy-logs.sh
```diff
@@ -0,0 +1,133 @@
...
+
+    # Generate a range of dates to make pattern matching easier
+    while [ $DAYS -ge 0 ]
+    do
```
rtfpessoa: For consistency would keep the do in the same line.
```suggestion
    while [ $DAYS -ge 0 ]; do
```

.editorconfig
pauloribeiro-codacy: As it is now, I don't think there will be a collision between this configuration file and the one for the linter. If there is, I will adapt the `.remarkrc` file.

docs/configuration/extract-codacy-logs.sh
```diff
@@ -0,0 +1,133 @@
...
+        LOGS_DATE=$(date -v -$DAYS\d "+%Y-%m-%d")
```
rtfpessoa: What is this command supposed to do? Does not seem to work on my machine (Ubuntu 18.04)

docs/configuration/logging.md
rtfpessoa: That is a good point. I prefer to deal with that in the script.
I will add a comment above.

docs/configuration/extract-codacy-logs.sh
```diff
@@ -0,0 +1,133 @@
...
+[ -z "$NAMESPACE" ] && usage
+
+CURRENT_DATE_TIME=$(date "+%Y%m%d-%H%M%S")
+LOGS_DIR="$(mktemp -d)/codacy_logs_$CURRENT_DATE_TIME"
+
+echo "Starting log files extraction"
+
+# Check if kubectl is available
+echo "Checking if kubectl is installed..."
+if ! kubectl version &>/dev/null; then
```
rtfpessoa: Since @paulopontesm brought this we should have a variable called `KUBECTL` and then check both `kubectl`  and `microk8s.kubectl` and assign it.

Then we can use the variable instead of the command.

You can check for the binaries with `which`.

docs/configuration/logging.md
rtfpessoa: If it does not work, we can use`https://raw.githubusercontent.com/codacy/chart/master/docs/configuration/extract-codacy-logs.sh`

docs/configuration/extract-codacy-logs.sh
afsmeira: It subtracts $DAYS from the generated date.

docs/configuration/extract-codacy-logs.sh
```diff
@@ -0,0 +1,155 @@
...
+    echo "Example: '$0 -n codacy -d 3' would retrieve the last 3 days of logs from the 'codacy' namespace of the kubernetes cluster"
+    exit 3
+}
+
+date_days_ago() {
+    OS=$(uname)
+    if [[ "$OS" == "Darwin" ]]; then
+        date -v -$1\d "+%Y-%m-%d"
+    elif [[ "$OS" == "Linux" ]]; then
+        date -d '-$1\day' "+%Y-%m-%d"
```
rtfpessoa:
```suggestion
        date -d '-1 day' "+%Y-%m-%d"
```

docs/configuration/extract-codacy-logs.sh
```diff
@@ -0,0 +1,155 @@
...
+    exit 4
+fi
+
+
+# Check if kubectl or microk8s.kubectl is available
+echo "Checking if kubectl is installed..."
+KUBECTL=$(which kubectl || which microk8s.kubectl)
+if [ $? -ne 0 ]; then
+    echo "kubectl not installed"
+    echo "Please install kubectl version specified in Codacy's documentation (see here - https://codacy.github.io/chart/#preparing-to-install-codacy) or the version used when installing your cluster"
```
paulopontesm: This section doesn't exist anymore https://codacy.github.io/chart/#preparing-to-install-codacy . There reference to kubectl was moved to this page https://codacy.github.io/chart/install/

docs/configuration/extract-codacy-logs.sh
```diff
@@ -0,0 +1,155 @@
...
+LOGS_DIR="$(mktemp -d)/codacy_logs_$CURRENT_DATE_TIME"
+
+echo "Starting log files extraction"
+
+# Check if zip is available
+echo "Checking if zip is installed..."
+if ! zip --version &>/dev/null; then
+    echo "zip not installed"
+    echo "Please install zip"
+    echo "Log files extraction failed, exiting..."
```
paulopontesm: can't we just have a single place to print this log?

docs/configuration/extract-codacy-logs.sh
```diff
@@ -0,0 +1,155 @@
+#!/usr/bin/env bash
+
+#### Helper functions ####
+
+usage()
+{
+    echo "Usage: $0 -n <NAMESPACE> [-d <DAYS_OF_LOGS>]"
+    echo "When using the -d flag, <DAYS_OF_LOGS> must be between 1 and 7"
```
paulopontesm: Why do we limit between 1 and 7 days?

We let customers configure the retention for this: `fluentdoperator.expirationDays` [here](https://github.com/codacy/chart) -> Also, should we explain in this doc how to configure this value also?

afsmeira: I don't think that over 7 days of logs are valuable when debugging issues. I do understand that we might be sending mixed messages but I think that in the script we shouldn't allow to extract more than 7 days, since the way we pattern match to find the correct days to extract is not cheap.

docs/configuration/extract-codacy-logs.sh
paulopontesm: I think the log makes sense, just add a method to print it or something so it isn't repeated everywhere

docs/configuration/extract-codacy-logs.sh
paulopontesm: hmm.. then we should say that at least in README.md on the home page.
Retention over 7 days is not recommended and

docs/configuration/extract-codacy-logs.sh
afsmeira: I forgot that one line. I'll fix it.

docs/configuration/extract-codacy-logs.sh
afsmeira: Ok, I'll add it there.

PR #317, opened 2020-03-22
Refactor logging

pauloribeiro-codacy:
Refactored the logging page, moving the section on how to obtain the logs from the server to its own page as suggested here [CY-1140](https://codacy.atlassian.net/browse/CY-1140).


docs/extra/logs-collect.md
```diff
@@ -0,0 +1,23 @@
+# Collecting logs for Support
+
+To help troubleshoot issues, obtain the logs from your Codacy instance and send them to Codacy's Support:
+
+1.  Collect logs from your Codacy instance using Fluentd:
+
+    ```bash
+    --set fluentdoperator.enabled=true
```
pauloribeiro-codacy: What is the full command to use here? Is it the one below?

`helm upgrade codacy codacy/codacy --set fluentdoperator.enabled=true`

If possible we should provide the full command here.

rtfpessoa: We should not have the full command here. It will get outdated pretty quickly.
I see your idea but it will outdated outdated.
You can put a link to the installation guide.

docs/extra/logs-collect.md
```diff
@@ -0,0 +1,23 @@
...
+
+    ```bash
+    bash <(curl -fsSL https://raw.githubusercontent.com/codacy/chart/master/docs/extra/extract-codacy-logs.sh) -n <namespace>
+    ```
+
+    The script compresses the logs in an archive file with the name `codacy_logs_<timestamp>.zip`.
+    
+    You can also [download the script](extract-codacy-logs.sh) to run it manually.
+
+3.  Send the archive file containing the logs to Codacy's support team for analysis.
```
rtfpessoa: Maybe add `support@codacy.com`

docs/extra/logs-collect.md
```diff
@@ -0,0 +1,23 @@
...
+
+To help troubleshoot issues, obtain the logs from your Codacy instance and send them to Codacy's Support:
+
+1.  Collect logs from your Codacy instance using Fluentd:
+
+    ```bash
+    --set fluentdoperator.enabled=true
+    ```
+
+    After executing this command the Fluentd daemonset will send the logs to MinIO, which is also installed by the Codacy chart.
```
rtfpessoa: Do we need to  have this here? 
Since you are separating I would just keep a link here pointing to the installation saying you need that to have logs.

mkdocs.yml
```diff
@@ -22,4 +22,27 @@ markdown_extensions:
     custom_checkbox: true
 - pymdownx.superfences
 - toc:
-    permalink: true
\ No newline at end of file
+    permalink: true
+
+# Table of contents
+nav:
+  - index.md
```
rtfpessoa: Why would you do this manually?

docs/configuration/logging.md
```diff
@@ -1,44 +1,19 @@
...
-1. [FluentBit](https://github.com/helm/charts/tree/master/stable/fluent-bit) as the agent
-2. [ElasticSearch + Kibana](https://github.com/elastic/helm-charts/tree/master/elasticsearch) for Aggregation and UI
-
-### Quick Installation
+To use the suggested stack, run the following commands to install the Fluent Bit and Elasticsearch charts on your cluster:
 
 ```bash
 helm repo add elastic https://helm.elastic.co
 helm install --name elasticsearch elastic/elasticsearch --namespace logging
 helm install --name fluent-bit stable/fluent-bit --namespace logging \
```
rtfpessoa: The things here to setup logs are the one in the other page.
I do not even understand the use case for one of our clients to have elasticsearch

docs/extra/logs-collect.md
afsmeira: And this field is now true by default, so I doubt that any client needs to change it at this point.

docs/extra/logs-collect.md
afsmeira: I don't know if we have other documentation for this process, but it might be possible that the log file is too big for email and that the client needs to agree on a way to send us the logs. Not sure if we should add that here.

docs/extra/logs-collect.md
```diff
@@ -0,0 +1,23 @@
...
+
+2.  Download the logs by running the following command locally, replacing `<namespace>` with the namespace in which Codacy was installed:
+
+    ```bash
+    bash <(curl -fsSL https://raw.githubusercontent.com/codacy/chart/master/docs/extra/extract-codacy-logs.sh) -n <namespace>
```
afsmeira: The command now has another option I forgot to document. You can use the `-d <DAYS_NUM>` flag to extract just the last `<DAYS_NUM>` of logs. This flag is optional, defaults to 7 days and should have a value between 1 and 7. If you run the command without any arguments it will show you the usage, where this is documented.

docs/extra/logs-collect.md
pauloribeiro-codacy: Both are good suggestions.

I'm [figuring out](https://codacy.slack.com/archives/CQESXTUHW/p1585045849010900) what is the usual procedure for customers to send bigger files to support.

mkdocs.yml
pauloribeiro-codacy: This is needed to be able to control the information architecture - the order of the entries on the sidebar table of contents, as well as the names of the "parent folders" aggregating subpages.

This is important because otherwise the pages will appear in alphabetical order on the sidebar, instead of a logical order that makes sense for the users.

docs/extra/logs-collect.md
pauloribeiro-codacy: Indeed, the REAME.md file states that `fluentdoperator.enable` is set to True by default, with a log expiration date of 7 days. This is great because it means I can simplify the instructions. :smile:

docs/configuration/logging.md
paulopontesm: Lets delete this page then @pauloribeiro-codacy if a customer asks for it we add it back

docs/extra/logs-collect.md
pauloribeiro-codacy: Indeed, it felt like extraneous detail.

Since `fluentdoperator.enabled` is already true by default, I removed this entire instruction.

docs/extra/logs-collect.md
pauloribeiro-codacy: Yes, and since this is useful to reduce the size of the .zip file, I included this as a suggestion.

docs/configuration/logging.md
```diff
@@ -1,44 +1,19 @@
...
 
-This page contains two different sections with different targets:
+There are different solutions for you to aggregate logs in your cluster for day to day operation.
 
-- Aggregate logs for day to day opperation
-- Collect logs [for support](#collect-logs-for-support)
+Overall, all solutions follow the same model: you have an **agent** on each node to collect and forward the logs, an **aggregator** to persist the logs, and a **UI** that allows you to analyze the logs.
 
-## Aggregation of logs
+One of the simplest stacks that you could use is:
```
rtfpessoa: Its cool to have options but I still see no use case for the second. Actually if you use the second you have no way to get the logs to support

docs/extra/logs-collect.md
```diff
@@ -0,0 +1,23 @@
...
+    ```bash
+    bash <(curl -fsSL https://raw.githubusercontent.com/codacy/chart/master/docs/extra/extract-codacy-logs.sh) \
+        -n <namespace> -d <days>
+    ```
+
+    You can also download the script [extract-codacy-logs.sh](extract-codacy-logs.sh) to run it manually.
+
+2.  Send the compressed logs to Codacy's support team at [support@codacy.com](mailto:support@codacy.com) for analysis.
+
+    If the file is too big, please upload the file to either a cloud storage service such as [Google Drive](https://www.google.com/drive/) or to a file transfer service such as [WeTransfer](http://www.wetransfer.com/) and send us the link to the file instead.
```
rtfpessoa: I would say that having this wetransfer here is very unprofessional

docs/configuration/logging.md
pauloribeiro-codacy: Done!

docs/extra/logs-collect.md
pauloribeiro-codacy: I agree that mentioning 3rd party services here is not optimal, but it's the way that Support currently recommends sending logs. We can revisit this at a later time, however.

PR #325, opened 2020-03-26
feature: Add GitHub Apps app id and private key to config CY-1228

carlospinho04:


codacy/values.yaml
```diff
@@ -23,6 +23,14 @@ global:
     workerManager:
       url: "codacy-worker-manager"
 
+  github: {}
+    # if you set this provider configuration, you must save it on the admin/integrations ui.
+    # enabled: "false"
+    # app:
+    #  appId: "1234"
+    #  TODO: Add link to documentation
```
pedrocodacy: Remove this and create a task to track that its done when the docs are available

codacy/values.yaml
```diff
@@ -31,6 +39,10 @@ global:
     # protocol: "https"
     # disableSSL: "false"
     # isPrivateMode: "true"
+    # app:
+    #  appId: "1234"
+    #  TODO: Add link to documentation
```
pedrocodacy: Remove this and create a task to track that its done when the docs are available

codacy/values-production.yaml
```diff
@@ -7,6 +15,10 @@ global:
     # protocol: "https"
     # disableSSL: "false"
     # isPrivateMode: "true"
+    # app:
+    #   appId: "1234"
+    #   TODO: Add link to documentation
```
joaocosta-codacy: Remove this and create a task to track that its done when the docs are available

codacy/values-production.yaml
```diff
@@ -1,4 +1,12 @@
 global:
+  github: {}
+    # if you set this provider configuration, you must save it on the admin/integrations ui.
+    # enabled: "false"
+    # app:
+    #   appId: "1234"
+    #   TODO: Add link to documentation
```
joaocosta-codacy: Remove this and create a task to track that its done when the docs are available

.circleci/config.yml
```diff
@@ -57,7 +57,7 @@ references:
     NUM_NODES: 5
     HELM_REPOSITORY: codacy-unstable
     CODACY_URL: http://dev.k8s.dev.codacy.org
-    HELM_INSTALL_EXTRA_FLAGS: --set codacy-api.config.github.clientName=codacy-k8s-selfhosted-dev
+    HELM_INSTALL_EXTRA_FLAGS: --set codacy-api.config.github.clientName=codacy-k8s-selfhosted-dev --set global.github.app.appId=57855 --set global.github.app.privateKey=${GITHUB_PRIVATE_KEY_DEV} --set global.githubEnterprise.app.appId=2 --set global.githubEnterprise.app.privateKey=${GITHUB_ENTERPRISE_PRIVATE_KEY_DEV}
```
joaocosta-codacy: --set global.github.app.appId=57855
--set global.githubEnterprise.app.appId=2

do these numbers have to be hardcoded here? can't we generate some random numbers in-place?

carlospinho04: We ca't generate the values, this are real app ids, what i can do is adding them as env vars if you want

joaocosta-codacy: Got it. Yes I think it would be best to have those as env vars as well

paulopontesm:
Do these need to be updated?

https://github.com/codacy/chart/blob/master/docs/configuration/git-providers/github-enterprise.md

https://github.com/codacy/chart/blob/master/docs/configuration/git-providers/github-enterprise.md

carlospinho04:
Yes, we need to update it, we have a task to do it

PR #326, opened 2020-03-26
Update mkdocs.yml

mrfyda:
Changing name to follow mkdocs convention. "Site name can only contain letters, numbers, underscores, hyphens and forward-slashes. The regular expression we test against is '^[a-zA-Z0-9_\-/]+$'."

This is required for codacy/docs#5 where we're trying to include this repo into the future docs of Codacy.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -1,5 +1,5 @@
 # Project information
-site_name: codacy/chart
+site_name: Chart
 site_url: https://codacy.github.io/chart/
 
 # Configuration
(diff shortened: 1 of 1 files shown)


mkdocs.yml
```diff
@@ -1,5 +1,5 @@
 # Project information
-site_name: "Codacy Chart"
+site_name: codacy/chart
```
pauloribeiro-codacy: Since this string appears on the documentation some customers are already using, I suggest we change it to something more user-friendly:

```suggestion
site_name: Chart
```

image

PR #333, opened 2020-03-30
Fix extra flags

h314to:


h314to:
Try using eval outside of the makefile. Still convoluted, but at least you should be able to expand the variables.

.circleci/config.yml
```diff
@@ -165,15 +165,15 @@ references:
       - <<: *update_versions
       - deploy:
           name: Install Codacy
-          command: make -e -C .doks/ deploy_to_doks_from_local_dir
+          command: eval make -e -C .doks/ deploy_to_doks_from_local_dir HELM_INSTALL_EXTRA_FLAGS="$EXTRA_FLAGS"
```
paulopontesm: shouldn't this eval be inside the makefile for the EXTRA_FLAGS?

I wont remember to do eval everywhere

DReigada:
I might have a solution for this that does the eval inside the Makefile. I'm testing it

DReigada:
Superseded by: #335

PR #334, opened 2020-03-31
fix: added missing config tag

joaocosta-codacy:
This config field was missing from the yaml.
I caught this while testing helm 3, which was complaining about the fact that `workerResources` was not a valid field. I went to check `worker-manager` and found out this discrepancy.

### DIFF
--- a/codacy/values-microk8s.yaml
+++ b/codacy/values-microk8s.yaml
@@ -58,6 +58,7 @@ worker-manager:
   resources:
     limits:
       memory: 500Mi
+  config:
     workerResources:
       limits:
         memory: "2Gi"


joaocosta-codacy:
You are right in saying that this is helm3 specific.
If this config is missing in the values file, while using helm3, we get the following error:
`Error: UPGRADE FAILED: error validating "": error validating data: [ValidationError(Deployment.spec.template.spec.containers[0].resources): unknown field "pluginResources" in io.k8s.api.core.v1.ResourceRequirements, ValidationError(Deployment.spec.template.spec.containers[0].resources): unknown field "workerResources" in io.k8s.api.core.v1.ResourceRequirements, ValidationError(Deployment.spec.template.spec.containers[0].resources): unknown field "workers" in io.k8s.api.core.v1.ResourceRequirements]`

PR #337, opened 2020-04-01
fix: nfsprovision claim size

paulopontesm:


codacy/values-production.yaml
```diff
@@ -196,9 +196,12 @@ listener:
     host: <--- codacy-db-host --->
     service:
       port: 5432
+  # These configurations are only used if the `nfsserverprovisioner` is enabled
   persistence:
     claim:
-      size: 140Gi
+      # This is the size of the claim that the component will get from the provisioner (if used)
+      # Therefore this value needs to be smaller than `nfsserverprovisioner.persistence.size`
+      size: 190Gi
```
bmbferreira, line 204: shouldn't we do it like this? is it even possible? 😅 

`size: {{ sub Values.nfsserverprovisioner.persistence.size 10 }}`

codacy/values-microk8s.yaml
```diff
@@ -33,9 +33,12 @@ listener:
   resources:
     limits:
       cpu: 500m
+  # These configurations are only used if the `nfsserverprovisioner` is enabled
   persistence:
     claim:
-      size: 50Gi
+      # This is the size of the claim that the component will get from the provisioner (if used)
+      # Therefore this value needs to be smaller than `nfsserverprovisioner.persistence.size`
+      size: 90Gi
```
andreaTP, line 41: do we have those values summed up in the hw requirement documentation?

codacy/values-production.yaml
andreaTP: I see the reasoning, having a margin of 10 Gb isn't excessive?

codacy/values-microk8s.yaml
paulopontesm: we are just asking for the 200Gi

PR #339, opened 2020-04-05
Edit and improve monitoring

pauloribeiro-codacy:
I did a general edit on the monitoring configuration page to simplify the instructions better and improve the readability.

@h314to, please see my comments below as I still have a few questions that I would like to clarify.


docs/configuration/monitoring.md
```diff
@@ -1,77 +1,72 @@
...
-```yaml
-global:
-  codacy:
-    crow:
-      url: "http://codacy.example.com/monitoring"
-```
+    ```yaml
+    username: codacy
+    password: C0dacy123
+    ```
```
pauloribeiro-codacy, line 22: These default values don't appear on any of the `*.yaml` files provided by us. Is this correct?

docs/configuration/monitoring.md
```diff
@@ -1,77 +1,72 @@
...
 
-### Configuring your credentials
+Follow the steps below to configure Crow and change the default configurations:
 
-We highly recommend that you define a custom password for the Crow installation. You can do this either through the `values.yaml` file or through a `--set` parameter during the Helm installation process. This parameter can be configured as follows:
+1.  Set the `global.codacy.crow.url` value in your `values.yaml` file so that Crow correctly generates anchor links to your projects. For example:
```
pauloribeiro-codacy, line 26: To simplify the sequence of instructions, I preferred the method of editing the `.yaml` file over using the `--set` parameter on a Helm upgrade.

docs/configuration/monitoring.md
```diff
@@ -1,77 +1,72 @@
...
 
-* Through a `--set` parameter:
+    ```yaml
+    global:
+      codacy:
+        crow:
+          url: "http://codacy.example.com/monitoring"
+    ```
```
pauloribeiro-codacy, line 33: I was expecting to find this configuration at the top of the `values.yaml` file, but it's not there:

```yaml
global:
  codacy:
    url: "http://localhost:9000"
    backendUrl: "http://localhost:9000"
    api:
      url: "codacy-api"
    portal:
      url: "codacy-portal"
    hotspots:
      url: "codacy-hotspots-api"
    engine:
      url: "codacy-engine"
    engineApi:
      url: "codacy-engine"
    listener:
      url: "codacy-listener"
    activities:
      url: "codacy-activities"
    core:
      url: "codacy-core"
    remoteProviderService:
      url: "codacy-remote-provider-service"
    workerManager:
      url: "codacy-worker-manager"
```

Also, further down the `.yaml` file we have a configuration section for Crow that includes a value for setting the PostgreSQL user name and password. Shouldn't the user configure these values as well at this time?

```yaml
crow:
  replicaCount: 1
  fullnameOverride: codacy-crow
  crowdb:
    create: false
    host: codacy-crowdb
    postgresqlUsername: "codacy"
    postgresqlDatabase: "crow"
    postgresqlPassword: "PLEASE_CHANGE_ME"
    service:
      port: 5432
    fullnameOverride: codacy-crowdb
  config:
    play:
      http:
        context: "/monitoring"
    passwordAuth:
      password: "PLEASE_CHANGE_ME"
```

docs/configuration/monitoring.md
```diff
@@ -1,77 +1,72 @@
...
-```
+2.  Set the `crow.config.passwordAuth.password` value in your `values.yaml` file to define a custom password for Crow:
 
-* Through the `values.yaml` file:
+    ```yaml
+    crow:
+      config:
+        passwordAuth:
+          password: <--- Crow password --->
+    ```
```
pauloribeiro-codacy, line 42: On this part of the `.yaml` file we have a configuration section for Crow that includes a value for setting the PostgreSQL user name and password. Shouldn't the user configure these values as well at this time?

```yaml
crow:
  replicaCount: 1
  fullnameOverride: codacy-crow
  crowdb:
    create: false
    host: codacy-crowdb
    postgresqlUsername: "codacy"
    postgresqlDatabase: "crow"
    postgresqlPassword: "PLEASE_CHANGE_ME"
    service:
      port: 5432
    fullnameOverride: codacy-crowdb
  config:
    play:
      http:
        context: "/monitoring"
    passwordAuth:
      password: "PLEASE_CHANGE_ME"
```

docs/configuration/monitoring.md
h314to: Yes, it's in the source: https://bitbucket.org/qamine/crow/src/master/.helm/crow/values.yaml#lines-122

docs/configuration/monitoring.md
h314to: I don't think it should be there because it is not part of the main setup. Maybe crow's url shouldn't be global because it is not used outside of it's chart afaik.

docs/configuration/monitoring.md
h314to: Yes, using a file instead of a flag makes redeployment easier. I would suggest using a values file just for crow to make this section self-contained.

docs/configuration/monitoring.md
h314to: Yes, that's a good idea.

docs/configuration/monitoring.md
pauloribeiro-codacy: I see, this would be an extra `values-crow.yaml` that could be appended to the other values files in a `helm upgrade` command, as we're doing for other settings.

Who would be the best person to implement this change?

docs/configuration/monitoring.md
pauloribeiro-codacy: Added explicit instructions for this (https://github.com/codacy/chart/pull/339/commits/90c801b1f8afbcf48a628e9181cbb22640cd8e2c).

docs/configuration/monitoring.md
pauloribeiro-codacy: Splitting the Crow configurations into a separate .yaml file was also suggested yesterday by Caxaria and Hélio while doing a test (https://codacy.atlassian.net/browse/CY-1170).

Do you think you can do this split and I'll update the documentation accordingly, @h314to?

docs/configuration/monitoring.md
pauloribeiro-codacy: I think I understand this better now. These settings could be in the dedicated .yaml file just for Crow, right?

docs/configuration/monitoring.md
```diff
@@ -1,77 +1,78 @@
...
 
-We highly recommend that you define a custom password for the Crow installation. You can do this either through the `values.yaml` file or through a `--set` parameter during the Helm installation process. This parameter can be configured as follows:
+1.  Set the user name, password, and hostname to access the Crow PostgreSQL database [already created here](../requirements.md#preparing-postgresql-for-codacy):
 
-* Through a `--set` parameter:
+    ```yaml
+    crowdb:
+      postgresqlUsername: <--- codacy-db-username --->
+      postgresqlPassword: <--- codacy-db-password --->
+      host: <--- codacy-db-host --->
```
pauloribeiro-codacy, line 32: @h314to is there any chance these Crow DB settings could be moved to the `values-production.yaml` file, at the end of the DB configurations for all the other components? Something like this:

```yaml
  [...]
  jobsdb:
    create: false
    postgresqlUsername: <--- codacy-db-username --->
    postgresqlDatabase: jobs # You need to create the DB manually
    postgresqlPassword: <--- codacy-db-password --->
    service:
      port: 5432
  crowdb:
    create: false
    postgresqlUsername: <--- codacy-db-username --->
    postgresqlDatabase: jobs # You need to create the DB manually
    postgresqlPassword: <--- codacy-db-password --->
    service:
      port: 5432
```

This would make it easier for the administrator to configure these settings in one go while editing the `values-production.yaml`, most probably by doing a search and replace to define these values all at once.

codacy/values-crow.yaml
```diff
@@ -0,0 +1,22 @@
+global:
```
pauloribeiro-codacy, line 1: @h314to I moved all Crow configurations to this dedicated .yaml file. Also check the instructions where I ask the user to run the `helm upgrade [...] --values values-crow.yaml` command to use values set here.

codacy/values-crow.yaml
```diff
@@ -0,0 +1,22 @@
...
+  codacy:
+    crow:
+      url: <--- "http://codacy.example.com/monitoring" ---> # Crow URL
+crow:
```
pauloribeiro-codacy, line 5: Does this URL really need to include the path `/monitoring`? Since the same path is also configured on `config.play.http.context` maybe it could be removed from here. My concern is that since the users will edit this value, they may accidentally change or remove the path and something could break.

docs/configuration/monitoring.md
```diff
@@ -1,77 +1,61 @@
...
 
-```yaml
-helm upgrade (...options used to install codacy...) --set crow.config.passwordAuth.password=<--- crow password --->
-```
+    -   Set the correct hostname of your Codacy instance so that Crow correctly generates anchor links to your projects
+    -   Set the username, password, and hostname to access the Crow PostgreSQL database [already created here](../requirements.md#preparing-postgresql-for-codacy)
+    -   Set the password for accessing Crow
 
-* Through the `values.yaml` file:
+3.  Apply this configuration by performing a Helm upgrade. To do so append `--values values-crow.yaml` to the command [used to install Codacy](../index.md#2-installing-codacy):
```
pauloribeiro-codacy, line 30: @h314to What do you think of this (hopefully more straightforward) way of configuring Crow?

docs/configuration/monitoring.md
pauloribeiro-codacy: I implemented this change on commit https://github.com/codacy/chart/pull/339/commits/c0892513ab7c9711d71e3952279e0ea7d9f2a67d

docs/configuration/monitoring.md
```diff
@@ -1,77 +1,61 @@
 # Monitoring
 
-Currently two monitoring solutions are supported:
+Currently, we support two monitoring solutions:
 
-* [**Crow**](#setting-up-monitoring-using-crow) - A simple, lightweight, and built-in monitoring solution.
-* [**Prometheus + Grafana + Loki**](#setting-up-monitoring-using-grafana-prometheus-and-loki) - A comprehensive third-party monitoring solution, recommended for more advanced usage.
+-   **[Crow](#setting-up-monitoring-using-crow):** A simple, lightweight, and built-in monitoring solution.
```
paulopontesm, line 5: Should we mention that comes enabled by default?

docs/configuration/monitoring.md
paulopontesm: Since we are enabling this by default and it should just be configured in the `values-production.yaml`, should we have this here?

codacy/values-crow.yaml
pauloribeiro-codacy: This was fixed on https://github.com/codacy/chart/pull/371.

docs/configuration/monitoring.md
pauloribeiro-codacy: Having Crow enabled by default is much better (and simpler). I'm going to propose the changes. :+1:

docs/configuration/monitoring.md
pauloribeiro-codacy: No, and now I understand why at first look the Crow configuration seemed to be completely wrong! :sweat_smile:

PR #340, opened 2020-04-06
fix: Improve the commands to install microk8s

andreaTP:
- remove unnecessary command concatenation (e.g. ` && \`) -> better to fail at a specific point in case
 - use snap to install helm (we already do it for microk8s good to be consistent)
 - wait for microk8s only when needed

### DIFF
--- a/docs/infrastructure/microk8s-quickstart.md
+++ b/docs/infrastructure/microk8s-quickstart.md
@@ -33,8 +33,8 @@ Install MicroK8s on the machine:
 2.  Install MicroK8s from the `1.15/stable` channel:
 
     ```bash
-    sudo snap install microk8s --classic --channel=1.15/stable && \
-    sudo usermod -a -G microk8s $USER  && \
+    sudo snap install microk8s --classic --channel=1.15/stable
+    sudo usermod -a -G microk8s $USER
     sudo su - $USER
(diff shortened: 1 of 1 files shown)


pauloribeiro-codacy:
@andreaTP, I'm sorry that I only saw this pull request now.

Isn't the `&&` operator meant to ensure that the next chained command [is only executed if the previous one ended successfully](https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Lists)?

This would prevent further commands from being executed if anything goes wrong in the sequence of commands, and as such I think it's preferable to include the `&&` operator.

Let me know if you think this makes sense.

andreaTP:
@pauloribeiro-codacy thanks for getting back on this!

You are correct about the `&&` operator and it's semantics, this said there are 2 elements that make me think removing it is a better choice in this case:
 - debuggability -> if something fails I clearly recognize which step failed
 - is widespread -> e.g. https://snapcraft.io/docs/installing-snap-on-ubuntu you see:
```bash
sudo apt update
sudo apt install snapd
```
and not:
```bash
sudo apt update && sudo apt install snapd
```

In this case also we are not performing "disrupting" commands, and we are not going to cause much problems or damage if something fails in the middle.

PR #343, opened 2020-04-07
fix: get changelog for new dependencies

joaocosta-codacy:


.circleci/getChangelogs.sh
```diff
@@ -32,23 +32,46 @@ function cloneRepository() {
...
+function getDependencyRepoUrl() {
+    echo $(yq r $1 dependencies -j | jq ".[] | select(.name==\"$2\").git" | sed "s/\"//g")
+}
+
 rm -rf ./changelogs
 mkdir changelogs
 prepareChangelogMarkdown
 
-LATEST_TAG="stable"
+LATEST_TAG="latest"
```
joaocosta-codacy, line 65: this caused the latest release to have too many tags in the changelog. 
latest is the correct tag to use

.circleci/getChangelogs.sh
```diff
@@ -32,23 +32,46 @@ function cloneRepository() {
     old_version=$2
     new_version=$3
     repository_url=$4
-    echo "[OK] $project_name($repository_url) : $old_version -> $new_version"
+    echo "[OK] checkout: $project_name($repository_url)"
     git clone "$repository_url" "$project_name" --quiet
 }
 
+function getOldDependencyVersion(){
```
joaocosta-codacy, line 39: If there is a new component introduced in the requirements file, it will not have an old version.

this has happened with codacy-ingress in the release it was introduced, where the changelog had no info about it.

.circleci/getChangelogs.sh
```diff
@@ -32,23 +32,46 @@ function cloneRepository() {
     old_version=$2
     new_version=$3
     repository_url=$4
-    echo "[OK] $project_name($repository_url) : $old_version -> $new_version"
+    echo "[OK] checkout: $project_name($repository_url)"
```
pauloribeiro-codacy, line 35: Should this line come only after the `git clone`, if it is successful?

.circleci/getChangelogs.sh
```diff
@@ -17,8 +17,8 @@ function getChangelog() {
     old_version=$2
     new_version=$3
     repository_url=$4
-    echo " * getting changelog ..."
```
pauloribeiro-codacy, line 20: Why are we dropping this line? I think it helps understand what's going on. :)

.circleci/getChangelogs.sh
```diff
@@ -32,23 +32,46 @@ function cloneRepository() {
...
-    echo "[OK] $project_name($repository_url) : $old_version -> $new_version"
+    echo "[OK] checkout: $project_name($repository_url)"
     git clone "$repository_url" "$project_name" --quiet
 }
 
+function getOldDependencyVersion(){
+    old_version=$(getDependencyVersion $OLD_LOCK_FILE $1)
+    if [ "$old_version" == "" ]; then
+        # this means this dependency is new as there is no old_version as per the requirements file
+        # therefore we fetch the oldest tag from the git remote
```
pauloribeiro-codacy, line 43: Does it make sense to fetch all commits since the very first tag? :thinking: I can be overkill to include what could be a very long commit history.

Perhaps we could simply include a line saying something like:

"Component X added as a new dependency"

It's just an idea. :)

.circleci/getChangelogs.sh
joaocosta-codacy: The purpose was to say that it found the dependency on the requirements file and is now checking out this repo.

.circleci/getChangelogs.sh
joaocosta-codacy: because we added 
`echo "  * $old_version -> $new_version"`
just below :)

.circleci/getChangelogs.sh
joaocosta-codacy: If that makes more sense from a changelog perspective, sure. What is your opinion?
I am happy to make that change :)

pauloribeiro-codacy: I'm just worried that we'll be fetching **a lot** of commit lines each time we add a new dependency. I really think it's more useful to someone going through the changelog to understand that a new dependency was added. If we could also include the version of the dependency that was added, that would be even better.

.circleci/getChangelogs.sh
pauloribeiro-codacy: Yes, but I think it's no longer easy to understand that this is being done in the context of fetching the changelog. :thinking:

.circleci/getChangelogs.sh
paulopontesm: I agree with @pauloribeiro-codacy .. we should just fetch the oldest tag and do a print saying that a new component was added.
Next release will have more details.

.circleci/getChangelogs.sh
joaocosta-codacy: sorted

.circleci/getChangelogs.sh
joaocosta-codacy: Agreed. I actually implemented this without checking with @pauloribeiro-codacy :) I agree that it could introduce a lot of stuff. As @paulopontesm said, the next release will have more details in the changelog.

.circleci/getChangelogs.sh
```diff
@@ -17,8 +17,8 @@ function getChangelog() {
     old_version=$2
     new_version=$3
     repository_url=$4
-    echo " * getting changelog ..."
     cd "$project_name"
+    echo "  * $old_version -> $new_version"
```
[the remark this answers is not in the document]
joaocosta-codacy: I have actually combined the commit messages we discussed earlier. I am dismissing this suggestion.

.circleci/getChangelogs.sh
```diff
@@ -32,31 +45,37 @@ function cloneRepository() {
...
-
-LATEST_TAG="stable"
-git cat-file blob "$LATEST_TAG":"$NEW_LOCK_FILE" > "$OLD_LOCK_FILE"
-DEPENDENCIES=$(yq r $OLD_LOCK_FILE dependencies -j | jq ".[].name" | sed "s/\"//g")
+function handleNewDependency() {
+    echo "New dependency $1($2): $3"
+    appendToChangelog "" # add blank line in changelog.md
+    appendToChangelog "## $1([$2]())"
+    appendToChangelog "### $3"
+    appendToChangelog "* This dependency was introduced at this version."
```
pauloribeiro-codacy, line 57:
```suggestion
    appendToChangelog "* This dependency was introduced in this version."
```

:nerd_face:

.circleci/getChangelogs.sh
```diff
@@ -1,28 +1,41 @@
...
     new_version=$3
     repository_url=$4
-    echo " * getting changelog ..."
     cd "$project_name"
+    echo "Getting changelog: $old_version -> $new_version"
     git fetch --all --quiet
     git checkout tags/"$old_version" --quiet
-    echo "## $project_name($repository_url)" >> "../changelogs/changelog.md"
-    gitchangelog "$old_version" "$new_version" >> "../changelogs/changelog.md"
+    appendToChangelog "## $project_name([$repository_url]())"
```
pauloribeiro-codacy, line 36: What do you think of adding this space between the project name and the repository URL for easier readability?

```suggestion
    appendToChangelog "## $project_name ([$repository_url]())"
```

.circleci/getChangelogs.sh
```diff
@@ -32,31 +45,37 @@ function cloneRepository() {
...
-mkdir changelogs
-prepareChangelogMarkdown
-
-LATEST_TAG="stable"
-git cat-file blob "$LATEST_TAG":"$NEW_LOCK_FILE" > "$OLD_LOCK_FILE"
-DEPENDENCIES=$(yq r $OLD_LOCK_FILE dependencies -j | jq ".[].name" | sed "s/\"//g")
+function handleNewDependency() {
+    echo "New dependency $1($2): $3"
+    appendToChangelog "" # add blank line in changelog.md
+    appendToChangelog "## $1([$2]())"
```
pauloribeiro-codacy, line 55: Same thing here:

```suggestion
    appendToChangelog "## $1 ([$2]())"
```

PR #344, opened 2020-04-08
feature: Add resolveDbsIps configuration for worker manager

bmbferreira:


bmbferreira:
> Since this PR is adding a flag that we don't plan to continue using I would prefer to avoid adding complexity

@joaocosta-codacy @pauloribeiro-codacy what do you prefer? the configuration here, having it disabled by default on the component (https://bitbucket.org/qamine/worker-manager/pull-requests/791) or both?

pauloribeiro-codacy:
> > Since this PR is adding a flag that we don't plan to continue using I would prefer to avoid adding complexity
> 
> @joaocosta-codacy @pauloribeiro-codacy what do you prefer? the configuration here, having it disabled by default on the component (https://bitbucket.org/qamine/worker-manager/pull-requests/791) or both?

From my limited understanding of this topic, it seems to me that:

* The PR [fix: do not resolve dbs ips by default on worker startup](https://bitbucket.org/qamine/worker-manager/pull-requests/791/fix-do-not-resolve-dbs-ips-by-default-on) already fixes the issue I was "complaining about" :slightly_smiling_face: in the PR [feature: Add ssm and config to configmap to avoid resolving DB IPS on startup](https://bitbucket.org/qamine/worker-manager/pull-requests/785/feature-add-ssm-and-config-to-configmap-to/diff#comment-144094068) by setting the parameter to `false`.
* I believe that having the parameter set to `false` should be the recommended setting and the one that is less prone to giving troubleshooting headaches to customers.

As such, I agree with @andreaTP in the sense that the configuration should not be needed on the chart side as long as it is already disabled by default on the component. :+1:

README.md
```diff
@@ -195,6 +195,7 @@ The following parameters are specific to each Codacy component.
 | `codacy-api.metrics.serviceMonitor.enabled`         | Create the ServiceMonitor resource type to be read by prometheus operator.                          | `false`                                      |
 | `codacy-api.metrics.grafana_dashboards.enabled`     | Create the ConfigMap with the dashboard of this component. Can be imported through grafana sidecar. | `false`                                      |
 | `worker-manager.config.workers.dedicatedMax`        | TBD                                                                                                 | `2`                                          |
+| `worker-manager.config.workers.resolveDbsIps`       | Resolve Database's ips at workers start-up (to avoid DNS issues)                                    | `false`                                      |
```
pauloribeiro-codacy, line 198: From the standpoint of a layman, the way the description is written makes it seem that `true` is actually the option that resolves the hostname to an IP address every time the workers start, as one would expect.

It was only by reading the description of this issue [CY-1309](https://codacy.atlassian.net/browse/CY-1309) that I understood why it works exactly the opposite way. :slightly_smiling_face: 

I think that unless someone knows the internal architecture of these components it's hard to explain how this setting works and what is the impact of having it set to either `true` or `false`.

bmbferreira: 👍 

Ok so I'll close this PR, do the change in the worker manager configuration instead and not expose this on the chart

PR #346, opened 2020-04-13
docs: Update docs for GitHub Apps

mrfyda:
Depends on #345

PR #348, opened 2020-04-13
feature: Move GitHub self-hosted configuration to file [CY-1351]

rtfpessoa:
Depends on https://bitbucket.org/qamine/codacy-website/pull-requests/3255/feature-move-github-self-hosted/diff

Depends on #346 346


pauloribeiro-codacy:
@rtfpessoa, let's collaborate on the documentation for the several Git providers on this separate PR https://github.com/codacy/chart/pull/350, as I'm already starting to work there on improving these instructions.

Cc: @mrfyda

rtfpessoa:
@pauloribeiro-codacy can we do it one provider at a time? I would prefer to have smaller scopes.
But if you think it is important to do all at the same time or that it helps you in any way we can do like you said.

mrfyda:
Merged #346

pauloribeiro-codacy:
> @pauloribeiro-codacy can we do it one provider at a time? I would prefer to have smaller scopes.
> But if you think it is important to do all at the same time or that it helps you in any way we can do like you said.

Let's keep in sync. If we're working simultaneously on this, you can also open separate branches for each Git provider where you do both the "code" and doc changes, but you should merge what I've already got in my branch to yours to ensure we're in sync. But if the changes are small and don't take a lot of time to do, I think we can include all the doc changes in the other PR and we update everything in one go.

rtfpessoa:
@pauloribeiro-codacy I think it would be good if we wrote the docs in here too. I just want to close the website first and then we could finish this with docs.

pauloribeiro-codacy:
> @pauloribeiro-codacy I think it would be good if we wrote the docs in here too. I just want to close the website first and then we could finish this with docs.

I understand your idea now, and it makes sense.

I merged `master` into this branch, as it already includes big changes that are necessary for the updated docs.

docs/configuration/git-providers/github-cloud.md
```diff
@@ -1,28 +1,22 @@
...
 
-![GitHub Application](./images/github-token-retrieval.png)
+| Field                          | Value                     |
+| ------------------------------ | ------------------------- |
+| `global.github.enabled`        | `true`                    |
+| `global.github.app.name`       | See above how to generate |
+| `global.github.app.id`         | See above how to generate |
+| `global.github.app.privateKey` | See above how to generate |
+| `global.github.clientId`       | See above how to generate |
+| `global.github.clientSecret`   | See above how to generate |
```
pauloribeiro-codacy, line 20: Now, the Client ID and Client Secret are configured in the .yaml file together with the other configurations.

Please check @rtfpessoa .

docs/configuration/git-providers/github-enterprise.md
```diff
@@ -1,5 +1,11 @@
 # GitHub Enterprise
 
+## Create a GitHub App
```
rtfpessoa, line 3: I was not doing GitHubEnterprise in this pr.
If we want to do both I have to update https://bitbucket.org/qamine/codacy-website/pull-requests/3255/feature-move-github-self-hosted/diff

pauloribeiro-codacy: Ah, yes, I know. These changes already came from `master` (originally done here: https://github.com/codacy/chart/pull/346).

Can we simply assume that the docs on how to configure the Git providers are unstable for now and I'll stabilize them in the next couple of days?

pauloribeiro-codacy: Even so, the changes for GitHub Enterprise done on https://github.com/codacy/chart/pull/346 were just to take into account GitHub Apps authentication. They do not reflect the configuration of the Client ID and Client Secret on the .yaml file side yet.

rtfpessoa: When this is merged to master this is published?
If yes we should stop it.
Docs should be published for a release, or hotfix.
We should not have docs that do not match the real case.

docs/configuration/git-providers/github-cloud.md
rtfpessoa: Looks good

rtfpessoa, line 20: Since these are not generated, they are obtained from GitHub after creating the app. I think this way is more clear
```suggestion
| `global.github.clientId`       | See above how to obtain |
| `global.github.clientSecret`   | See above how to obtain |
```

rtfpessoa:
@pauloribeiro-codacy before we merge this I will do a manual sandbox test, to make sure it works fine. If that works it should be ready to go.

rtfpessoa:
I also forgot to add the breaking tag. Since this will not be backwards compatible.
Users that used the ui in k8s will have to manually move this to the file.

@mrfyda @Laracodacy @fjrdomingues  is this a concearn rn?

mrfyda:
afaik this is not live yet

docs/configuration/git-providers/github-enterprise.md
pauloribeiro-codacy: I agree with you @rtfpessoa.

To clarify:
* When you opened this PR, the documentation was heavily out of date and had not been updated to take into account GitHub Apps authentication yet.
* While this PR was already ongoing, @mrfyda updated the documentation  on `master` to take into account GitHub Apps (PR https://github.com/codacy/chart/pull/346).
* Because it's easier to make these updates on moving the configurations to .yaml files on top of a more up-to-date base documentation, I merged `master` to this branch (https://github.com/codacy/chart/pull/348/commits/c44972f71092222ec14d0c6c04d11b9b3d3fc8cd). Otherwise, we would have tricky conflicts when we finally merged this PR back to `master`...
* After this merge I worked on updating the documentation to take into account the new configurations being done directly in .yaml files instead of the interface.

Which means we should be ready to go. :slightly_smiling_face:

rtfpessoa:
This is not working in sandbox. I will have to keep digging what other configs are hammered in the website

pauloribeiro-codacy:
I did another merge from `master` to fix the existing conflicts.

rtfpessoa:
@pauloribeiro-codacy 

> I did another merge from master to fix the existing conflicts.
Ideally we should not do merges, because it ends up trashing the commit list and because it now does not allow for the rebase.

I am not sure why this repo does not allow merge commit.

PR #351, opened 2020-04-14
Provision external dbs for self-hosted

joaocosta-codacy:


.doks/doks-cluster/cluster.tf
```diff
@@ -5,15 +5,27 @@ resource "digitalocean_kubernetes_cluster" "codacy_k8s" {
 
   node_pool {
     name       = "codacy-doks-pool-${terraform.workspace}"
-    size       = var.node_type
-    node_count = var.num_nodes
+    size       = "s-2vcpu-2gb"
```
joaocosta-codacy, line 8: This was added to match the remote terraform state.

.doks/doks-cluster/cluster.tf
```diff
@@ -5,15 +5,27 @@ resource "digitalocean_kubernetes_cluster" "codacy_k8s" {
...
+    min_nodes = 0
+    max_nodes = 0
   }
 
   provisioner "local-exec" {
     command = "doctl kubernetes cluster kubeconfig save ${var.cluster_name} --set-current-context"
   }
 }
 
+resource "digitalocean_kubernetes_node_pool" "auto-scale-pool-01" {
```
joaocosta-codacy, line 20: This was added to match the remote terraform state

.circleci/config.yml
```diff
@@ -73,8 +75,9 @@ references:
     DO_TF_WORKSPACE: dev
     RELEASE_NAME: codacy-release
     NAMESPACE: codacy-release
-    K8S_VERSION: "1.15"
-    NUM_NODES: "5"
+    ENVIRONMENT: release
```
joaocosta-codacy, line 78: Used for pattern-matching with the templated values-db-<environment>.yaml file created by terraform

.circleci/config.yml
```diff
@@ -64,7 +65,8 @@ references:
     DO_TF_WORKSPACE: dev
     RELEASE_NAME: codacy-hourly
     NAMESPACE: codacy-hourly
-    K8S_VERSION: "1.14"
+    ENVIRONMENT: sandbox
```
joaocosta-codacy, line 68: Used for pattern-matching with the templated values-db-<environment>.yaml file created by terraform

.circleci/config.yml
```diff
@@ -53,8 +53,9 @@ references:
     DO_TF_WORKSPACE: dev
     RELEASE_NAME: codacy
     NAMESPACE: codacy
-    K8S_VERSION: "1.14"
-    NUM_NODES: "5"
+    ENVIRONMENT: dev
```
joaocosta-codacy, line 56: Used for pattern-matching with the templated values-db-<environment>.yaml file created by terraform

.doks/doks-cluster/databases.tf
```diff
@@ -0,0 +1,213 @@
...
+  cluster_id = digitalocean_database_cluster.postgres-dev.id
+  name       = "postgres-dev-pool-${each.key}"
+  mode       = var.connection_pool_mode
+  size       = var.connection_pool_size
+  db_name    = each.key
+  user       = var.postgres_default_admin
+   depends_on = [digitalocean_database_db.postgres_dev_db]
+}
+
+resource "digitalocean_database_connection_pool" "postgres-dev-pool-defaultdb" {
```
joaocosta-codacy, line 54: Provisioned separately as defaultdb comes created out of the box. Therefore, I could not reuse the foreach that i used above to create all dbs plus defaultdb

.doks/doks-cluster/locals.tf
```diff
@@ -0,0 +1,281 @@
...
+    postgresqlDatabase: postgres-dev-pool-jobs
+    postgresqlPassword: ${digitalocean_database_cluster.postgres-dev.password}
+    host: ${digitalocean_database_connection_pool.postgres-dev-pool["jobs"].private_host}
+    service:
+      port: ${digitalocean_database_connection_pool.postgres-dev-pool["jobs"].port}
+activities:
+  activitiesdb:
+    create: false
+    postgresqlUsername: ${digitalocean_database_cluster.postgres-dev.user}
+    postgresqlDatabase: "activities?sslmode=require"
```
joaocosta-codacy, line 60: This db was failing to receive connections. The Activities pod was entering an `Error` state because the connection was non-ssl. Then I tried (with the amazing help of @h314to kudos!) to connect to the connection pool with SSL and there were still other errors occurring. The only way in which we managed to establish a connection with SSL enabled was to not use the connection pool, and to pass the ssl mode as part of the db name, which will get appended to the connection string that goes on the component's configmap.

.doks/doks-cluster/locals.tf
```diff
@@ -0,0 +1,281 @@
...
+    postgresqlPassword: ${digitalocean_database_cluster.postgres-dev.password}
+    host: ${digitalocean_database_cluster.postgres-dev.private_host}
+    service:
+      port: ${digitalocean_database_cluster.postgres-dev.port}
+hotspots-api:
+  hotspotsdb:
+    create: false
+    postgresqlUsername: ${digitalocean_database_cluster.postgres-dev.user}
+    postgresqlDatabase: "hotspots?sslmode=require"
```
joaocosta-codacy, line 69: See the comment on activitiesdb about sslmode being required.

.doks/doks-cluster/locals.tf
```diff
@@ -0,0 +1,281 @@
...
+    postgresqlDatabase: postgres-dev-pool-listener
+    postgresqlPassword: ${digitalocean_database_cluster.postgres-dev.password}
+    host: ${digitalocean_database_connection_pool.postgres-dev-pool["listener"].private_host}
+    service:
+      port: ${digitalocean_database_connection_pool.postgres-dev-pool["listener"].port}
+crow:
+  crowdb:
+    create: false
+    postgresqlUsername: ${digitalocean_database_cluster.postgres-dev.user}
+    postgresqlDatabase: "crow?sslmode=require"
```
joaocosta-codacy, line 87: See the comment on activitiesdb about sslmode being required.

.doks/doks-cluster/variables.tf
```diff
@@ -58,3 +62,42 @@ variable "cluster_name" {
...
+variable "postgres_instance_type" {
+  description = "The instance type for the postgres clusters."
+  type = string
+  default = "db-s-4vcpu-8gb"
+}
+
+variable "connection_pool_size" {
+  description = "Number connections for each connection pool."
+  type        = number
+  default     = 17
```
joaocosta-codacy, line 96: With the current db cluster size, this is the maximum number of connections we can have across all connection pools.
There is one edge case that I verified with the `engine` component that is important to keep in mind.
When rotating the deployment of the engine, the newly created pod fails to successfully connect to all dbs, due to insufficient available connections on the pools. This means that the deployment will be in a limbo state as the new `engine` pod never becomes ready.
This may happen with other components as they rotate, although I am not absolutely sure.
The easy solution is, of course, to bump up the instance type which will retroactively allow us to increase the conn. pool sizes. We just need to foot the bill in that case :)

docs/extra/database.md
```diff
@@ -45,7 +46,7 @@ This will dump the file with the `.sql.tar` extension into the `/tmp` folder.
 To restore a database, you can run a `pg_restore` command to consume the dump file and replicate the data onto Postgres:
 
 ```bash
-PGPASSWORD=$DB_PASSWORD pg_restore -h $HOSTNAME -U $DB_USER -d $DB -F t /tmp/$DB.sql.tar --clean --if-exists --create
+PGPASSWORD=$DB_PASSWORD pg_restore -h $DEST_HOSTNAME -p $DEST_HOSTPORT -U $DB_USER -j 8 -d $db -n public --clean $db.dump
```
joaocosta-codacy, line 49: `-j 8`  means that pg_restore will run 8 jobs (processes) in parallel

.circleci/config.yml
```diff
@@ -53,8 +53,9 @@ references:
...
+    K8S_VERSION: 1.14
+    NUM_NODES: 5
```
paulopontesm, line 58: Is this used? You removed this from the makefile. Why was this removed?

.circleci/config.yml
paulopontesm: I don't love this dependency.

.doks/doks-cluster/databases.tf
```diff
@@ -0,0 +1,213 @@
+######################
+# cluster-dev
```
paulopontesm, line 2: Can-t we just reuse this template 3 times with a variable that recieves the name of the DB?

.doks/doks-cluster/locals.tf
```diff
@@ -0,0 +1,281 @@
+######################
+# values_dbs_dev.yaml
+######################
+
+locals{
+  template_values_dbs_dev = <<EOF
+global:
```
paulopontesm, line 7: can't we just `sed` this on the `values-production.yaml`? 

The user,password, host and port should be the same for all dbs, right?

joaocosta-codacy: I see your point but we use the .doks/values.yaml for our digitalocean environments. There are different settings there such are component resources or persistence sizes on values-production.yaml.

As for not using `sed`, I am changing that now. :)

.doks/doks-cluster/databases.tf
joaocosta-codacy: We probably can. It did not occur to me 👍 
I did that for the dbs but not for the clusters themselves. Changing it now.

.doks/doks-cluster/locals.tf
joaocosta-codacy: Actually, now i remember why!
If you use a local_exec, it will be saved in the state. Which means that the next time you run the terraform apply, that specific resource will not be created. Therefore, you will not have your templated file in the local filesystem on the build directory.

.doks/doks-cluster/databases.tf
joaocosta-codacy: Although that will force the replacement (destruction and creation) of the environments. Meaning, re-importing the dbs :-(

.circleci/config.yml
joaocosta-codacy: It was just converted from string to int

.circleci/config.yml
joaocosta-codacy: The main purpose of this is so that the correct values file for the specific environment is picked up in the makefile while installing codacy. 
The db hosts as well as the user and password for that host will change per environment so I don't think we can just use one file for everything. How do you think it would be best to implement this?

.doks/doks-cluster/databases.tf
```diff
@@ -0,0 +1,67 @@
+######################
```
paulopontesm, line 1: It's better, less repeated stuff. 
Still, I would put this in a different directory than "doks-cluster" (it isn't a doks cluster).

After that you can just re-use it inside circleci wth `terraform apply database/ -var 'name=codacy-postgres-$ENVIRONMENT'`

With this I can make changes to the database without having to do in all of them at the same time

.circleci/config.yml
paulopontesm: Where is this used?

joaocosta-codacy: This was used when we were trying to have separate clusters per environment. But that's no longer the case so I vote for removing K8S_VERSION. The terraform files will control which k8s version to use in the cluster.
NUM_NODES was used when we did not have an autoscaling pool. We can drop it as well.

.doks/doks-cluster/databases.tf
joaocosta-codacy: Everything digital ocean is now on the same workspace. Which means that first I will move the db related resources to a separate workspace, since moving the files to a separate folder will cause some resource definitions (k8s cluster for example) to be absent. Meaning, terraform would just want to delete our k8s cluster.

The main problem now is that, because we decided to go with a non-declarative approach (with nested for loops to reduce the amount of repetition), we can not apply individual configurations to the db clusters. Say that you want to have a better instance type for the release environment. Or that you want to bump sandbox to postgres11 for testing. We can not do that.
Unless what you are meaning is to be able to provision db clusters a-la-carte, to create a brand new environment for example. That's valuable but is completely out of the scope of the original ticket which was just to bump postgres on digital ocean to 10. :'(

codacy/values-production.yaml
```diff
@@ -94,47 +94,47 @@ global:
     postgresqlPassword: <--- codacy-db-password --->
     host: <--- codacy-db-host --->
     service:
-      port: 5432
+      port: <--- codacy-db-port --->
```
paulopontesm, line 97: Do you need to do this? Isn't the port the default one? Can't we just "not replace" this one?

.do/db-cluster/Makefile
```diff
@@ -0,0 +1,53 @@
+# terraform setup
```
paulopontesm, line 1: Why do we need a specific makefile for this one? Isn't it most of the stuff here just duplicated from the other one?
Can we have a single one under .doks?

.do/db-cluster/Makefile
```diff
@@ -0,0 +1,53 @@
...
+	$(TF) workspace select $(WORKSPACE)
+
+.PHONY: terraform_logistics
+terraform_logistics:
+	$(TF) apply -parallelism=$(TF_PARALLELISM) -var-file=$(VARS_FILE) -input=false --auto-approve
+	$(TF) output -json > terraform_outputs.json
+
+.PHONY: replace_db_values
+replace_db_values:
+	$(eval host=$(shell cat terraform_outputs.json | jq -r '.db_cluster_hosts.value[] | .host | select(contains("$(ENVIRONMENT)"))'))
```
paulopontesm, line 47: Why the `select(contains("$(ENVIRONMENT)"))'` ? Isn't the name fixed?

paulopontesm: Using the `terraform output` also seems "cleaner"

https://www.terraform.io/docs/commands/output.html

paulopontesm, line 47: Instead of the eval I think it would be better just to do a single command with `&&`. It stops if you get an error.

```
host=$(shell terraform output  .... ) && \
port=$(shell terraform output  .... ) && \
```

.do/db-cluster/databases.tf
```diff
@@ -0,0 +1,34 @@
+resource "digitalocean_database_cluster" "postgres" {
```
paulopontesm, line 1: rename the file to "database.tf"

paulopontesm: non plural

.do/db-cluster/databases.tf
```diff
@@ -0,0 +1,34 @@
...
+  for_each = local.environments
```
paulopontesm, line 2: We don't need `for` here. The same template but you just recieve the env name by env variable.

What you call  `each.key` in the entire document can be called `var.environment`

.do/db-cluster/locals.tf
```diff
@@ -0,0 +1,3 @@
+locals{
```
paulopontesm, line 1: I think we don't need this document

.do/db-cluster/outputs.tf
```diff
@@ -0,0 +1,27 @@
+output "db_cluster_hosts" {
```
paulopontesm, line 1: Same for this. We don't need the "fors" just the value.
Then we can get the values with `terraform output ...`
https://www.terraform.io/docs/commands/output.html

.do/db-cluster/databases.tf
```diff
@@ -0,0 +1,34 @@
...
+  name = "codacy-doks-cluster-postgres${var.postgres_version}-${each.key}"
```
paulopontesm, line 3: I am not sure if we should put the postgres.version in the name.
Doesn't the DB get replaced when we set this to "11"? Instead of just updating "in place"?

.do/k8s-cluster/Makefile
```diff
@@ -3,14 +3,14 @@ TF_RC ?= ./terraformrc
 TF_EXTRA_OPTS ?= TF_IN_AUTOMATION=1 AWS_SDK_LOAD_CONFIG=1 TF_CLI_CONFIG_FILE=$(TF_RC)
 TF_PARALLELISM ?= 30
 TF ?= $(TF_EXTRA_OPTS) terraform
+NAMESPACE ?= codacy
 WORKSPACE ?= ${DO_TF_WORKSPACE}
 VARS_FILE ?= deploy.tfvars
 NGINX_INGRESS_VERSION ?= 1.17.1
 # variables
-K8S_VERSION ?= 1.15
+K8S_VERSION ?= 1.14
```
paulopontesm, line 11: why?

.circleci/config.yml
paulopontesm: Ok.. I think it makes sense to have it there

PR #352, opened 2020-04-14
feature: Add bitbucket cloud configs and documentation CY-1376

bmbferreira:
Do not merge this yet (wait for portal and website)


docs/configuration/git-providers/bitbucket.md
```diff
@@ -0,0 +1,35 @@
+# Bitbucket
+
+Since Bitbucket uses OAuth, the first step is to register a consumer for Codacy on Bitbucket side.
+
+## Create a consumer
+OAuth needs a key and secret, together these are know as an OAuth consumer. You can create a consumer on any existing individual or team account. To create a consumer, do the following:
```
prendi, line 6: known as

README.md
```diff
@@ -116,6 +116,9 @@ Global parameters apply to all sub-charts and make it easier to configure resour
 | `global.gitlabEnterprise.hostname`              | Hostname of gitlabEnterprise instance                                                                        | `nil`           |
 | `global.gitlabEnterprise.protocol`              | Protocol of gitlabEnterprise instance                                                                        | `nil`           |
 | `global.gitlabEnterprise.port`                  | Port of gitlabEnterprise instance                                                                            | `nil`           |
+| `global.bitbucket.enabled`                      | Enable bitbucket                                                                                             | `nil`           |
```
pauloribeiro-codacy, line 119: I know this is also not consistent for the other Git providers, but here we should capitalize the name of the Git provider correctly:

```suggestion
| `global.bitbucket.enabled`                      | Enable Bitbucket                                                                                             | `nil`           |
```

codacy/values.yaml
```diff
@@ -51,6 +51,11 @@ global:
     # protocol: "https"
     # port: 443
 
+  bitbucket: {}
+    # enabled: "false"
+    # key: ""
+    # secret: ""
+
```
pauloribeiro-codacy, line 58: This is also done for the other Git providers, but do we need to have these settings repeated on both the `values.yaml` and `values-production.yaml` files?

On the [installation instructions](https://codacy.github.io/chart/#2-installing-codacy) we're asking the customers to configure their settings on the `values-production.yaml` file. So I think it's less confusing to have the settings only in one place.

docs/configuration/git-providers/bitbucket.md
```diff
@@ -0,0 +1,35 @@
...
+        ![Bitbucket Consumer Configuration](./images/bitbucket-consumer-configuration.png)
+     1. Add the following permissions: 
+     ![Bitbucket Consumer Permissions](./images/bitbucket-consumer-permissions.png)
+     1. Click Save. 
+        The system generates a key and a secret for you.
+     1. Toggle the consumer name to see the generated Key and Secret value for your consumer.
+     ![Bitbucket Consumer Key and Secret](./images/bitbucket-consumer-key-and-secret.png)
+
+## Enable Bitbucket and set the key and the secret
+Set your configuration values for Bitbucket on the `values.yaml` file.
```
pauloribeiro-codacy, line 22: This should be the `values-production.yaml` file instead:

```suggestion
Set your configuration values for Bitbucket on the `values-production.yaml` file.
```

docs/configuration/git-providers/bitbucket.md
```diff
@@ -0,0 +1,35 @@
...
+1. Copy the secret from Bitbucket and set it in the `secret`field.
+
+Example:
+```yaml
+global:
+  bitbucket:
+    enabled: "true"
+    key: "12345232123"
+    secret: "54321abcxdawj4bfsdh3"
+```
```
pauloribeiro-codacy, line 35: We also need to run a command to update the settings on the cluster, something like:

```
helm upgrade (...options used to install Codacy...) \
              --values values-production.yaml
```

docs/configuration/git-providers/bitbucket.md
bmbferreira: notice that we have `values.yaml` in every place 😨

docs/configuration/git-providers/bitbucket.md
bmbferreira: should we repeat this command for every provider instructions?

codacy/values.yaml
bmbferreira: hmm then I don't know the purpose of having these values.yaml if we always use values-production.yaml. Is the values.yaml just for reference? @h314to @paulopontesm @joaocosta-codacy

pauloribeiro-codacy: The `values.yaml` file is a default configuration file, that can be overwritten by other `.yaml` files that are passed explicitly as parameters to the `helm` command:

https://helm.sh/docs/chart_template_guide/values_files/

So the logic being followed so far is that we use the `values.yaml` the set the defaults but all the configurations that must be configured by the user are specified in the `values-production.yaml` (and `values-microk8s.yaml` as well if the user is installing on MicroK8s).

But let's wait for someone from the Reliability team to confirm this.

docs/configuration/git-providers/bitbucket.md
pauloribeiro-codacy: I know, I'm already fixing this in another PR for the Monitoring configuration and will also do another full review when both the Inquisition and Jonah's squads finish their current work on the Git providers. So don't worry about the other occurrences. :+1:

docs/configuration/git-providers/bitbucket.md
pauloribeiro-codacy: I will make sure this is consistent for all the Git providers when both the Inquisition and Jonah's finish their current work. So, let's ensure the instructions are correct for Bitbucket Cloud in this PR and I'll take care of the other providers.

bmbferreira:
Do not merge this yet. We're still testing on the sandbox.

docs/configuration/git-providers/bitbucket.md
```diff
@@ -0,0 +1,40 @@
...
+## Enable Bitbucket and set the key and the secret
+Set your configuration values for Bitbucket on the `values-production.yaml` file.
+
+1. Enable Bitbucket by setting `enable` to `"true"`.
+1. Copy the key from Bitbucket and set it in the `key` field.
+1. Copy the secret from Bitbucket and set it in the `secret`field.  
+   Example:
+   ```yaml
+   global:
+   bitbucket:
```
mrfyda, line 30:
```suggestion
     bitbucket:
```

PR #353, opened 2020-04-14
fix: GitHub App docs

pauloribeiro-codacy:
@mrfyda, I fixed:

* The new page on how to create a GitHub App was missing from the mkdocs.yaml navigation
* The link to an image was broken
* The .yaml file that we are asking customers to edit is the `values-production.yaml` and not the `values.yaml`


codacy/values.yaml
```diff
@@ -27,8 +27,8 @@ global:
     # if you set this provider configuration, you must save it on the admin/integrations ui.
     # enabled: "false"
     # app:
-    #  id: "1234"
     #  name: "codacy"
+    #  id: "1234"
```
pauloribeiro-codacy, line 31: @mrfyda I changed the order of this setting because it makes more sense to have the name at the top, just like on the GitHub interface for creating GitHub Apps.

PR #359, opened 2020-04-17
fix: Add missing quote on helm extra arguments

machadoit:


mrfyda:
😬

machadoit:
It seems that it could be introduced during conflicts resolution

PR #360, opened 2020-04-17
docs: Apply user feedback CY-1170

pauloribeiro-codacy:
Applying feedback to the documentation resulting from the user test with @caxaria on 2020-04-09.

Notes available here:
https://codacy.atlassian.net/browse/https://codacy.atlassian.net/browse/CY-1170

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -76,7 +76,13 @@ Install Codacy on an existing cluster using our Helm chart:
 6.  Add Codacy's chart repository to your Helm client and install the Codacy chart using the file `values-production.yaml` created previously.
 
     !!! important
-        **If you are using MicroK8s** don't forget to use the file `values-microk8s.yaml` together with the file `values-production.yaml` as [described here](infrastructure/microk8s-quickstart.md#notes-on-installing-codacy). To do this, uncomment the last line before running the command below.
+        **If you are using MicroK8s** you must use the file `values-microk8s.yaml` together with the file `values-production.yaml`.
+        
+        Use `wget` to download the extra file and uncomment the last line before running the `helm upgrade` command below:
+
+        ```bash
+        wget https://raw.githubusercontent.com/codacy/chart/master/codacy/values-microk8s.yaml
+        ```
 
     ```bash
     helm repo add codacy-stable https://charts.codacy.com/stable/
--- a/docs/index.md
+++ b/docs/index.md
@@ -93,33 +93,31 @@ Install Codacy on an existing cluster using our Helm chart:
                  # --values values-microk8s.yaml
     ```
 
-7.  By now all the Codacy pods should be starting in the cluster. Run the following command **and wait for all the pods to have the status Running** to verify this:
+7.  By now all the Codacy pods except `codacy-crow` (which can be [configured later](configuration/monitoring.md#setting-up-monitoring-using-crow)) should be starting in the cluster.
+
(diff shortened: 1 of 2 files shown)


docs/infrastructure/microk8s-quickstart.md
```diff
@@ -41,70 +41,93 @@ Install MicroK8s on the machine:
...
-1.  Enable the following plugins on MicroK8s:
+1.  Configure MicroK8s to allow privileged containers:
 
     ```bash
     sudo mkdir -p /var/snap/microk8s/current/args
     sudo echo "--allow-privileged=true" >> /var/snap/microk8s/current/args/kube-apiserver
     microk8s.status --wait-ready
+    ```
+
+2.  Enable the following MicroK8s addons:
```
pauloribeiro-codacy, line 59: @paulopontesm, what do you think about having the `microk8s.status --wait-ready` after enabling each MicroK8s addon? Hopefully this will help avoid the race condition that caused some addons to not be enabled correctly.

pauloribeiro-codacy: I just realized I'm reverting a change that @andreaTP made in https://github.com/codacy/chart/pull/340/commits/cbf8bf709ba696d5f5f30393f6eb34e23fc08029#diff-eaa6927aebd91897b0dfeac25e036d13L71.

However, I think it's safer to include the "waiting time" between enabling each addon as I've seen the glitch happen both on my local Ubuntu virtual machine as in AWS EC2 instances.

paulopontesm: I think we should add it. 
If it helps with flakiness I believe it's ok. Also, it's not a shady command and should be quite straight forward for anyone that it following the guide to understand what it does

paulopontesm: Although, it looks like the --wait sometimes is asynchronous and it doesn't wait for the resources of the addon to be created, it just waits for the reply that the instruction was received successfully. DNS seems to be the one that generates more problems. Should we create a disclaimer or something saying that this could happen? Just re-run...

docs/index.md
```diff
@@ -87,33 +93,31 @@ Install Codacy on an existing cluster using our Helm chart:
                  # --values values-microk8s.yaml
     ```
 
-7.  By now all the Codacy pods should be starting in the cluster. Run the following command **and wait for all the pods to have the status Running** to verify this:
+7.  By now all the Codacy pods except `codacy-crow` (which can be [configured later](configuration/monitoring.md#setting-up-monitoring-using-crow)) should be starting in the cluster.
```
paulopontesm, line 96: The crowdb should be configured in the values-production.yaml and start right away.

@joaocosta-codacy I think we have a bug here.

joaocosta-codacy: I am looking into it.

joaocosta-codacy: https://github.com/codacy/chart/pull/371

docs/infrastructure/microk8s-quickstart.md
pauloribeiro-codacy: Indeed it makes sense to add the disclaimer / note.

I was also thinking of moving the DNS addon to be activated last, but since [it is required by other addons](https://microk8s.io/docs/addon-dns) (probably the Ingress needs it?) I just left it as the first one.

docs/index.md
pauloribeiro-codacy: Wow, this is great! I'll revert these changes, then. :+1:

PR #369, opened 2020-04-20
Fix number of replicas for recommended MicroK8s

pauloribeiro-codacy:
Fixing the number of replicas as discussed during the "Training on K8s" meeting.

### DIFF
--- a/docs/requirements.md
+++ b/docs/requirements.md
@@ -55,7 +55,7 @@ The resources recommended on the following table are based on our experience and
 | Kubernetes<br/>Medium Installation (default) | 2                      | 4                        | 8 vCPUs<br/>16 GB RAM    | 20 vCPUs<br/>40 GB RAM    | **32 vCPUs<br/>64 GB RAM**    |
 | Kubernetes<br/>Big Installation              | 2+                     | 10+                      | 8+ vCPUs<br/>16+ GB RAM  | 50+ vCPUs<br/>100+ GB RAM | **60+ vCPUs<br/>110+ GB RAM** |
 | MicroK8s<br/>Minimum                         | 1                      | 2                        | 5.5 vCPUs<br/>9.5 GB RAM | 10 vCPUs<br/>20 GB RAM    | **16 vCPUs<br/>32 GB RAM**    |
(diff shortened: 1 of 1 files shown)


h314to:
Hum... most of the components only have one replica. Only a couple of them have replicas. Maybe 1+ is more accurate than 2+, because the latter would imply that all of them have at least two replicas.

pauloribeiro-codacy:
> Hum... most of the components only have one replica. Only a couple of them have two replicas, really. Maybe 1+ is more accurate than 2+, because the latter would imply that all of them have at least two replicas.

Thanks @h314to, I understand what you mean. Either way, then maybe the "Platform resources" shouldn't double, then? Should we ask only for 5.5+ vCPUs and 9.5+ GB RAM?

The values for this last row of the table should be revised, because as they are now we are asking two times for the "extra 1.5 vCPU and 1.5 GB memory for the MicroK8s platform itself running on the same machine as the Platform".

@paulopontesm, can you also have a look?

paulopontesm:
> > Hum... most of the components only have one replica. Only a couple of them have two replicas, really. Maybe 1+ is more accurate than 2+, because the latter would imply that all of them have at least two replicas.
> 
> Thanks @h314to, I understand what you mean. Either way, then maybe the "Platform resources" shouldn't double, then? Should we ask only for 5.5+ vCPUs and 9.5+ GB RAM?
> 
> The values for this last row of the table should be revised, because as they are now we are asking two times for the "extra 1.5 vCPU and 1.5 GB memory for the MicroK8s platform itself running on the same machine as the Platform".
> 
> @paulopontesm, can you also have a look?

We are putting 2 replicas on the website and engine. That is 1 CPUs + 1.5 Gb extra

pauloribeiro-codacy:
> > > Hum... most of the components only have one replica. Only a couple of them have two replicas, really. Maybe 1+ is more accurate than 2+, because the latter would imply that all of them have at least two replicas.
> > 
> > 
> > Thanks @h314to, I understand what you mean. Either way, then maybe the "Platform resources" shouldn't double, then? Should we ask only for 5.5+ vCPUs and 9.5+ GB RAM?
> > The values for this last row of the table should be revised, because as they are now we are asking two times for the "extra 1.5 vCPU and 1.5 GB memory for the MicroK8s platform itself running on the same machine as the Platform".
> > @paulopontesm, can you also have a look?
> 
> We are putting 2 replicas on the website and engine. That is 1 CPUs + 1.5 Gb extra

@paulopontesm please confirm if the values in my last commit are the ones you had in mind. :white_check_mark:

PR #376, opened 2020-04-25
Review configuration of integrations

pauloribeiro-codacy:
Reviewed the documentation on how to configure the Git providers and STMP email server, ensuring consistency across providers, fixing small bugs, and improving overall usability of the instructions.

Please check the individual commit messages for the rationale behind some of the changes.


docs/configuration/integrations/gitlab-enterprise.md
```diff
@@ -1,35 +1,56 @@
...
+        https://<codacy hostname>/add/addProvider/GitLab
+        https://<codacy hostname>/add/addService/GitLab
+        https://<codacy hostname>/add/addPermissions/GitLab
+        ```
+
+    -   **Scopes:** Enable the scopes:
+    
+        - `api`
+        - `read_user`
+        - `read_repository`
```
pauloribeiro-codacy, line 24: @jllopes Are these the correct configurations? They do not match what we had for Doplicated:

https://support.codacy.com/hc/en-us/articles/207994975-Configuring-GitLab-Enterprise-with-Codacy-Self-hosted

I also noticed that the configurations that existed before this merge used to match the Doplicated ones:

https://github.com/codacy/chart/commit/0dd82a1dcbab7e58a8a57633dda06f61d82f22c5#diff-dc8c06a3b4b1ffac2c62498bb26e0739R3

jllopes: Hey @pauloribeiro-codacy. Good catch, the URL's should have GitLabEnterprise and not GitLab, and the same for the yaml suggestion

jllopes: Also, it seems that the OpenID scope is also required

docs/configuration/integrations/gitlab-enterprise.md
```diff
@@ -1,35 +1,57 @@
...
-    ```
-    https://codacy.example.com/login/GitLab
-    https://codacy.example.com/add/addPermissions/GitLab
-    https://codacy.example.com/add/addProvider/GitLab
-    https://codacy.example.com/add/addService/GitLab
-    ```
+    -   **Redirect URI:** Copy the URLs below, updating the HTTP protocol and hostname to the correct values of your Codacy instance. This field is case sensitive.
+
+        ```text
+        https://<codacy hostname>/login/GitLabEnterprise
```
rtfpessoa, line 14: I think our usage of placeholders vs examples is a bit mixed, do we have any reason for that @pauloribeiro-codacy ?

Also, in this case if the hostname is a placefolder the protocol and the port should be too, probably.

docs/configuration/integrations/email.md
```diff
@@ -2,25 +2,36 @@
...
-            password: "a-s!r0ng-pVsSwORG"
-            # port: 25
+      enabled: "true"
+      replyTo: "notifications@mycompany.com" # Reply-to field on sent emails
+      smtp:
+        hostname: "smtp.mycompany.com" # Hostname of your SMTP server
+        protocol: "smtps" # SMTP protocol to use, either smtps or smtp
+        username: "my-smtp-username" # Username to authenticate on your SMTP server
+        password: "a-s!r0ng-pVsSwORG" # Password to authenticate on your SMTP server
+        port: 465 # Port of your SMTP server
```
rtfpessoa, line 18: The idea for some fields to be commented was because they were optional.
Actually here username, password and port are optional depending on the smtp configuration of the server

docs/configuration/integrations/gitlab-cloud.md
```diff
@@ -1,35 +1,55 @@
 # GitLab Cloud
 
-## GitLab Application
-
 Follow the instructions below to set up the Codacy integration with GitLab Cloud:
 
-1.  Create a [new application](https://gitlab.com/profile/applications) pointing to your local Codacy deployment URL with `api`, `read user` and `read repository` scopes.
+1.  Open <https://gitlab.com/profile/applications>.
```
afsmeira, line 5: Will this be formatted as a link? Otherwise just convert it to one, it's simpler.

pauloribeiro-codacy: Yes, this is a bare URL in Markdown and is rendered like this:

image

(For the self-hosted providers I'm not formatting these URLs as anchor links because the user must enter the correct hostname of their instance, so clicking the link would only be confusing.)

docs/configuration/integrations/gitlab-enterprise.md
pauloribeiro-codacy: I was trying to be more explicit, but I agree with you that this is not consistent right now and may cause more confusion.

I will simplify these to examples, and will also sync the .yaml settings between the `values-production.yaml` and the documentation so that everything is more straightforward and easier to maintain.

docs/configuration/integrations/email.md
pauloribeiro-codacy: On the `values-production.yaml` only the port was commented out twice:

```
  email: {}
    # enabled: "false"
    # replyTo: "notifications@codacy.com"
    # smtp:
    #   protocol: "smtps" # smtps | smtp
    #   hostname: "smtp.codacy.com"
    #   username: "codacy"
    #   password: "a-s!r0ng-pVsSwORG"
    #   # port: 25
```

So I interpreted this as "if you omit the port, it will use port 25 by default".

But indeed, it's perfectly possible to have an SMTP server with no authentication, especially if it's internal and protected with a firewall.

Is there any difference between the two options below?

```
  email: {}
    # enabled: "false"
    # replyTo: "notifications@codacy.com"
    # smtp:
    #   protocol: "smtps" # smtps | smtp
    #   hostname: "smtp.codacy.com"
    #   # username: "codacy"
    #   # password: "a-s!r0ng-pVsSwORG"
    #   # port: 25
```

or

```
  email: {}
    # enabled: "false"
    # replyTo: "notifications@codacy.com"
    # smtp:
    #   protocol: "smtps" # smtps | smtp
    #   hostname: "smtp.codacy.com"
    #   username: ""
    #   password: ""
    #   port: ""
```

If Codacy interprets the empty values in the same way as if they are not defined, I would prefer to have the comments saying that the settings are optional, as this is more explicit:

```
  email: {}
    # enabled: "false"
    # replyTo: "notifications@codacy.com"
    # smtp:
    #   protocol: "smtps" # smtps | smtp
    #   hostname: "smtp.codacy.com"
    #   username: "" # Optional username to authenticate on your SMTP server
    #   password: "" # Optional password to authenticate on your SMTP server
    #   port: "" # Optional port of your SMTP server, the default is 25
```

rtfpessoa: You need to comment them to make them undefined.
It might work with helm if string is empty, but I would definitely not rely on that behaviour.

pauloribeiro-codacy: -hmmm- I see... It's better to include the "extra comment" then.

PR #377, opened 2020-04-26
Add networking requirements

pauloribeiro-codacy:
Added the networking requirements as identified in this issue:

https://codacy.atlassian.net/browse/DOCS-34

This was also identified while doing this user test:

https://codacy.atlassian.net/browse/CY-1170

### DIFF
--- a/docs/requirements.md
+++ b/docs/requirements.md
@@ -22,15 +22,16 @@ The cluster running Codacy must satisfy the following requirements:
 
 The cluster must be configured to accept and establish connections on the following ports:
 
-|          | Service      | Protocol/Port | Notes                                                        |
-| -------- | ------------ | ------------- | ------------------------------------------------------------ |
-| Inbound  | SSH          | TCP/22        | Optional, to access the infrastructure remotely.             |
-| Inbound  | HTTP         | TCP/80        |                                                              |
-| Inbound  | HTTPS        | TCP/443       |                                                              |
-| Outbound | PostgreSQL   | TCP/5432      | Connection to the PostgreSQL DBMS                            |
-| Outbound | SMTP         | TCP/25        | Connection to your SMTP server                               |
-| Outbound | SMTPS        | TCP/465       | Connection to your SMTP server over TLS/SSL                  |
-| Outbound | Git provider | \*            | Connection to the ports required by your remote Git provider |
+|          | Service      | Protocol/Port | Notes                                                              |
+| -------- | ------------ | ------------- | ------------------------------------------------------------------ |
+| Inbound  | SSH          | TCP/22        | **MicroK8s only**, to access the infrastructure remotely.          |
+| Inbound  | HTTP         | TCP/80        |                                                                    |
+| Inbound  | HTTPS        | TCP/443       |                                                                    |
+| Outbound | PostgreSQL   | TCP/5432      | Connection to the PostgreSQL DBMS                                  |
+| Outbound | SMTP         | TCP/25        | Connection to your SMTP server                                     |
+| Outbound | SMTPS        | TCP/465       | Connection to your SMTP server over TLS/SSL                        |
+| Outbound | Docker Hub   | \*            | Connection to Docker Hub to download the required container images |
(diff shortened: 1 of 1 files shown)


docs/requirements.md
```diff
@@ -11,13 +11,27 @@ The next sections describe in detail how to set up these prerequisites.
...
+|          | Service      | Protocol/Port | Notes                                                        |
+| -------- | ------------ | ------------- | ------------------------------------------------------------ |
+| Inbound  | SSH          | TCP/22        | Optional, to access the infrastructure remotely.             |
+| Inbound  | HTTP         | TCP/80        |                                                              |
+| Inbound  | HTTPS        | TCP/443       |                                                              |
+| Outbound | PostgreSQL   | TCP/5432      | Connection to the PostgreSQL DBMS                            |
+| Outbound | SMTP         | TCP/25        | Connection to your SMTP server                               |
+| Outbound | SMTPS        | TCP/465       | Connection to your SMTP server over TLS/SSL                  |
+| Outbound | Git provider | \*            | Connection to the ports required by your remote Git provider |
+
```
pauloribeiro-codacy, line 34: I'm not sure if this information is enough for customers, or if we need to go into the detail of specifying which IP addresses / ranges are involved. :thinking: 

From my point of view we should have this simplified information for now and only add more detail if customers ask for it.

docs/requirements.md
```diff
@@ -11,13 +11,27 @@ The next sections describe in detail how to set up these prerequisites.
...
 -   Tiller, the server part of [Helm](https://v2.helm.sh/) **version 2.16**, must be installed in the cluster
 -   The [NGINX Ingress controller](https://github.com/kubernetes/ingress-nginx) must be installed and correctly set up in the cluster
 
+### Cluster networking requirements
+
+The cluster must be configured to accept and establish connections on the following ports:
+
+|          | Service      | Protocol/Port | Notes                                                        |
+| -------- | ------------ | ------------- | ------------------------------------------------------------ |
+| Inbound  | SSH          | TCP/22        | Optional, to access the infrastructure remotely.             |
```
paulopontesm, line 27: For "regular" kubernetes it doesn't make sense to put the 22 port. 
We only need kubectl correctly configured.

For microk8s we need the port 22.

docs/requirements.md
```diff
@@ -11,13 +11,27 @@ The next sections describe in detail how to set up these prerequisites.
...
+| Inbound  | HTTP         | TCP/80        |                                                              |
+| Inbound  | HTTPS        | TCP/443       |                                                              |
+| Outbound | PostgreSQL   | TCP/5432      | Connection to the PostgreSQL DBMS                            |
+| Outbound | SMTP         | TCP/25        | Connection to your SMTP server                               |
+| Outbound | SMTPS        | TCP/465       | Connection to your SMTP server over TLS/SSL                  |
+| Outbound | Git provider | \*            | Connection to the ports required by your remote Git provider |
```
rtfpessoa, line 33: Shouldn't we have dockerhub here too?

docs/requirements.md
pauloribeiro-codacy: Changing the notes to "**MicroK8s only**, to access the infrastructure remotely".

docs/requirements.md
pauloribeiro-codacy: We should. :+1: 

Unfortunately it seems it's not trivial to define the firewall rules to whitelist access to the Docker Hub. Docker ends up not providing the details (https://success.docker.com/article/is-there-a-whitelist-of-ips-used-by-docker-hub) and I also couldn't find a definitive answer online.

So I suggest that we add another row to the table with the aim of reminding the administrator about this and not so much to actually provide the values needed to configure a firewall.

rtfpessoa: I think its two different problems. We never had ips in the configuration.
They could probably whitelist the dns instead, at least that was what we provided before.

rtfpessoa: I tried to google about it and maybe it was never an option. Not sure why we had it there before.

PR #378, opened 2020-04-27
feature: Enable gitlab integration

jllopes:


.doks/values.yaml
```diff
@@ -29,6 +29,10 @@ global:
     protocol: "https"
     disableSSL: "true"
     isPrivateMode: "true"
+    
+  gitlab:
```
machadoit, line 33: It is also missing here https://github.com/codacy/chart/blob/master/codacy/values-production.yaml ?

bmbferreira:
is this enough? where are the credentials for the API?

jllopes:
> is this enough? where are the credentials for the API?

They are on the circleCI config.yml

PR #383, opened 2020-04-28
feature: Add codacy-spa CY-1359

DReigada:


DReigada:
Will open a PR from a normal named branch

PR #384, opened 2020-04-28
Review Git provider integration troubleshooting

pauloribeiro-codacy:
Centralized the troubleshooting information in a single documentation page to make it easier to link from the product and compacted the information that was repeated across Git providers to make the documentation easier to maintain.


docs/configuration/integrations/troubleshoot.md
```diff
@@ -0,0 +1,85 @@
...
+
+While trying to authenticate on Bitbucket Cloud you get the following error message:
+
+![Invalid client_id](images/bitbucket-invalid-client-id.png)
+
+This means that there is a mismatch in the OAuth consumer Client ID that Codacy is using to authenticate on Bitbucket Cloud.
+
+To solve this issue:
+
+1.  Make sure that the value of `key` in your `values-production.yaml` file is the same as the Key of the [Bitbucket OAuth consumer that you created](bitbucket-cloud.md#create-oauth)
```
pauloribeiro-codacy, line 79: @rtfpessoa please verify this. In this case the Client ID actually corresponds to the setting `global.bitbucket.key`, correct?

docs/configuration/integrations/troubleshoot.md
```diff
@@ -0,0 +1,85 @@
...
+
+## GitHub Cloud and GitHub Enterprise authentication {id="github"}
+
+### 404 error
+
+While trying to authenticate on GitHub you get the following error message:
+
+![Invalid client id](images/github-invalid-client-id.png)
+
+This means that there is a mismatch in the Client ID that Codacy is using to authenticate on GitHub.
```
rtfpessoa, line 15: Maybe here `This means` might be too strong. (now that I read it again :smile: )
We could probably say `This might mean`.

docs/configuration/integrations/troubleshoot.md
```diff
@@ -0,0 +1,85 @@
...
+
+To solve this issue:
+
+1.  Make sure that the value of `clientId` in your `values-production.yaml` file is the same as the Client ID of the [GitHub App that you created](github-app-create.md)
+2.  If the values were different, you need to re-execute the `helm upgrade` command as described for [GitHub Cloud](github-cloud.md) or [GitHub Enterprise](github-enterprise.md)
```
rtfpessoa, line 20: I do not think this is what you need.
We had a task (or at least talked about it) to evaluate reload of configs.

I think the last I heard you would need to take down the namespace. Not sure if a better command was introduced (i know its possible)

cc\ @paulopontesm @h314to @joaocosta-codacy

docs/configuration/integrations/troubleshoot.md
```diff
@@ -0,0 +1,85 @@
...
+While trying to authenticate on GitLab you get the following error message:
+
+![Invalid application id](images/gitlab-invalid-application-id.png)
+
+This means that there is a mismatch in the Application ID that Codacy is using to authenticate on GitLab.
+
+To solve this issue:
+
+1.  Make sure that the value of `clientId` in your `values-production.yaml` file is the same as the Application ID of the [GitLab Cloud](gitlab-cloud.md#create-application) or [GitLab Enterprise](gitlab-enterprise.md#create-application) application that you created
+2.  If the values were different, you need to re-execute the `helm upgrade` command as described for [GitLab Cloud](gitlab-cloud.md#configure) or [GitLab Enterprise](gitlab-enterprise.md#configure)
```
rtfpessoa, line 60: Same as above

docs/configuration/integrations/troubleshoot.md
rtfpessoa: :+1:

docs/configuration/integrations/troubleshoot.md
pauloribeiro-codacy: Indeed, now that you mention this, I'm not sure if a simple `helm upgrade` is enough to correctly force the reload of these configurations. :thinking: 

@heliocodacy has experience with this, maybe he can also help.

heliocodacy: If we change something on the values.yaml, it's recommended to use the flag --recreate-pods to ensure that the pods that need that configuration are recreated.

pauloribeiro-codacy: Yes, now I remember that in the test with Caxaria he did use this extra flag every time it was necessary to update the settings (we can check this in minute 47:12 of the [recording](https://codacy.slack.com/files/UTFLDNDN2/F0124RZULFK/mk8s_install_caxaria_-_part_1.mp4), for example).

rtfpessoa: Can we make that explicit?

docs/configuration/integrations/troubleshoot.md
```diff
@@ -0,0 +1,85 @@
...
+While trying to authenticate on GitLab you get the following error message:
+
+![Invalid application id](images/gitlab-invalid-application-id.png)
+
+This might mean that there is a mismatch in the Application ID that Codacy is using to authenticate on GitLab.
+
+To solve this issue:
+
+1.  Make sure that the value of `clientId` in your `values-production.yaml` file is the same as the Application ID of the [GitLab Cloud](gitlab-cloud.md#create-application) or [GitLab Enterprise](gitlab-enterprise.md#create-application) application that you created
+2.  If the values were different, you need to re-execute the `helm upgrade` command as described for [GitLab Cloud](gitlab-cloud.md#configure) or [GitLab Enterprise](gitlab-enterprise.md#configure)
```
afsmeira, line 60: I think there is something missing in this sentence. I would rephrase to "If the values were different change your configuration and then re-execute (...)". I feel that you should explicitly say the configuration needs to change. What do you think?

If you end up changing this, you'll have to do it for all providers.

pauloribeiro-codacy: When I wrote this I was thinking that the "Make sure" in the previous step already implied changing the configuration. But you're right, it doesn't hurt at all to be more explicit than implicit! :+1:

docs/configuration/integrations/troubleshoot.md
pauloribeiro-codacy: Yep, I'm actually adding this flag in each place where it is required.

PR #387, opened 2020-05-04
feature: Add secrets troubleshooting [CY-1336]

rtfpessoa:
Related to https://bitbucket.org/qamine/codacy-website/pull-requests/3379/wip-feature-move-secrets-value-to-logs-cy

### DIFF
--- a/docs/configuration/secrets/extract-codacy-secrets.sh
+++ b/docs/configuration/secrets/extract-codacy-secrets.sh
@@ -2,8 +2,9 @@
 
 usage()
 {
+    echo "Retrieves secrets used to encrypt sensitive data on the database from the Codacy logs."
     echo "Usage: $0 -n <NAMESPACE>"
-    echo "Example: '$0 -n codacy' would retrieve secrets from the logs from the 'codacy' namespace of the kubernetes cluster"
+    echo "Example: '$0 -n codacy' retrieves secrets from Codacy running on the cluster namespace 'codacy'"
     exit 3
 }
 
@@ -27,8 +28,8 @@ echo "Checking if kubectl is installed..."
 KUBECTL=$(which kubectl || which microk8s.kubectl)
 if [ $? -ne 0 ]; then
     echo "kubectl not installed"
-    echo "Please install kubectl version specified in Codacy's documentation (see here - https://codacy.github.io/chart/install/) or the version used when installing your cluster"
-    echo "To install kubectl see https://kubernetes.io/docs/tasks/tools/install-kubectl/ (or https://microk8s.io/docs/ if you are running a microk8s kubernetes cluster)"
+    echo "Please install the kubectl version specified in Codacy's documentation (https://codacy.github.io/chart/install/) or the version used when installing your cluster"
+    echo "To install kubectl see https://kubernetes.io/docs/tasks/tools/install-kubectl/, or https://microk8s.io/docs/ if you are running a MicroK8s kubernetes cluster"
     exit 4
 fi
 
@@ -42,7 +43,7 @@ fi
 
 read -p "Is '$KUBE_CTX' the correct kubernetes cluster for log extraction? (yes/[no]): " ANSWER
 if [[ ! "$ANSWER" =~ ^y(es)?$ ]]; then
-    echo "Please configure correctly your current kubernetes cluster"
+    echo "Configure your kubernetes cluster context to access the cluster where Codacy is running"
     exit 6
 fi
 
@@ -55,7 +56,7 @@ if [ $? -ne 0 ]; then
 fi
 
 DEPLOYMENT_LOGS_COMMAND="$KUBECTL -n $NAMESPACE logs deployment/codacy-api"
-SECRET_MESSAGES=$($DEPLOYMENT_LOGS_COMMAND | grep -i "Your database is using key")
+SECRET_MESSAGES=$($DEPLOYMENT_LOGS_COMMAND | grep -i "Sensitive data on the database will be encrypted using the secret")
 if [ $? -ne 0 ]; then
     echo "Failed to get the logs for the 'codacy-api' deployment, for namespace $NAMESPACE"
     echo "You can try running the command by hand ('$DEPLOYMENT_LOGS_COMMAND') or contacting support@codacy.com"
--- a/docs/configuration/secrets/extract-codacy-secrets.sh
+++ b/docs/configuration/secrets/extract-codacy-secrets.sh
@@ -1,0 +1,65 @@
+#!/usr/bin/env bash
+
+usage()
+{
+    echo "Usage: $0 -n <NAMESPACE>"
+    echo "Example: '$0 -n codacy' would retrieve secrets from the logs from the 'codacy' namespace of the kubernetes cluster"
+    exit 3
+}
+
+while getopts "n:" option; do
+    case ${option} in
+        n )
+            NAMESPACE=${OPTARG}
+            ;;
+        \? )
+            usage
+            ;;
+    esac
+done
+
+[ -z "$NAMESPACE" ] && usage
+
+echo "Starting secrets extraction"
+
+# Check if kubectl or microk8s.kubectl is available
+echo "Checking if kubectl is installed..."
+KUBECTL=$(which kubectl || which microk8s.kubectl)
+if [ $? -ne 0 ]; then
+    echo "kubectl not installed"
+    echo "Please install kubectl version specified in Codacy's documentation (see here - https://codacy.github.io/chart/install/) or the version used when installing your cluster"
+    echo "To install kubectl see https://kubernetes.io/docs/tasks/tools/install-kubectl/ (or https://microk8s.io/docs/ if you are running a microk8s kubernetes cluster)"
+    exit 4
+fi
+
+# Check current cluster context
+echo "Checking access to kubernetes cluster..."
+KUBE_CTX=$($KUBECTL config current-context)
+if [ $? -ne 0 ]; then
+    echo "No kubernetes cluster context configured"
+    exit 5
+fi
+
+read -p "Is '$KUBE_CTX' the correct kubernetes cluster for log extraction? (yes/[no]): " ANSWER
+if [[ ! "$ANSWER" =~ ^y(es)?$ ]]; then
+    echo "Please configure correctly your current kubernetes cluster"
+    exit 6
+fi
+
+echo "Checking if the deployment exists..."
+DEPLOYMENT_INFORMATION=$($KUBECTL -n $NAMESPACE get deployments | grep codacy-api)
+if [ $? -ne 0 ]; then
+    echo "Failed to get the 'codacy-api' deployment, for namespace $NAMESPACE"
+    echo "Are you sure you are in the right kubernetes cluster context?"
+    exit 7
+fi
+
+DEPLOYMENT_LOGS_COMMAND="$KUBECTL -n $NAMESPACE logs deployment/codacy-api"
+SECRET_MESSAGES=$($DEPLOYMENT_LOGS_COMMAND | grep -i "Your database is using key")
+if [ $? -ne 0 ]; then
+    echo "Failed to get the logs for the 'codacy-api' deployment, for namespace $NAMESPACE"
+    echo "You can try running the command by hand ('$DEPLOYMENT_LOGS_COMMAND') or contacting support@codacy.com"
+    exit 8
+fi
+
+echo "$SECRET_MESSAGES"
--- a/docs/configuration/secrets/troubleshoot.md
+++ b/docs/configuration/secrets/troubleshoot.md
@@ -6,26 +6,27 @@ If the information provided on this page is not enough to solve your issue, cont
 
 ## Lost/Changed initially generated secrets
 
-While trying to open Codacy,
-you were shown an error message stating that,
-the secret on the database and the one in your configuration file are different.
+When you open the Codacy UI, an error message states that the secret on the database and the one in your configuration file are different.
 
 ### Steps
 
-1. Get the latest logs from the API service
+1. Obtain the correct key used to encrypt sensitive data in the database from the Codacy logs.
 
-    a. Get logs and filter by the prefix of the message
+    ```bash
+    bash <(curl -fsSL https://raw.githubusercontent.com/codacy/chart/master/docs/configuration/secrets/extract-codacy-secrets.sh) \
+        -n <namespace>
+    ```
 
-        ```bash
-        bash <(curl -fsSL https://raw.githubusercontent.com/codacy/chart/master/docs/configuration/secrets/extract-codacy-secrets.sh) \
-            -n <namespace>
-        ```
+    You can also download the script [extract-codacy-secrets.sh](extract-codacy-secrets.sh) to run it manually.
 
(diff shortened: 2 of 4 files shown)


docs/configuration/secrets/troubleshoot.md
```diff
@@ -0,0 +1,34 @@
...
+### Steps
+
+1. Get the tail of the logs on the api service
+
+    a. Setup access to the cluster
+
+    b. Get logs and filter by the prefix of the message
+
+        ```sh
+        kubectl -n codacy logs deployment/codacy-api | grep -i "Your database is using key"
```
rtfpessoa, line 22: Should we do a script similar to the one for the logs for this?
It has the same cluster setup reqs.
@afsmeira @jllopes

docs/configuration/secrets/troubleshoot.md
```diff
@@ -0,0 +1,34 @@
...
+        ```
+        ```
+2. Copy the value and update it in your values yaml file
+
+3. Restart the application with the new configuration
+
+    ```bash
+    helm upgrade (...options used to install Codacy...) \
```
rtfpessoa, line 30: This line is starting to be repeated in too many places. I am starting to worry it will get outdated and cause issues.

docs/configuration/secrets/troubleshoot.md
```diff
@@ -0,0 +1,34 @@
...
+1. Get the tail of the logs on the api service
+
+    a. Setup access to the cluster
+
+    b. Get logs and filter by the prefix of the message
+
+        ```sh
+        kubectl -n codacy logs deployment/codacy-api | grep -i "Your database is using key"
+        ```
+        ```
```
jllopes, line 24: Extra closing ``` here

docs/configuration/secrets/troubleshoot.md
jllopes: Yeah, I think it should, since it makes the troubleshooting easier

docs/configuration/secrets/troubleshoot.md
jllopes: Maybe we should have a document for commands repeated in multiple places and just start linking into that document? @pauloribeiro-codacy

docs/configuration/secrets/troubleshoot.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Lost/Changed initially generated secrets
+
+While trying to open Codacy,
+you were shown an error message stating that,
+the secret on the database and the one in your configuration file are different.
+
+### Steps
+
+1. Get the tail of the logs on the api service
```
afsmeira, line 15: Three comments:

- Why mention tail specifically? I think "Get the latest logs from the api service" would be more appropriate. 
- Not sure if we are using "api" or "API".
- Guarantee that we are always calling 'api service' to the website backend.

docs/configuration/secrets/troubleshoot.md
afsmeira: Yes, I agree.

docs/configuration/secrets/troubleshoot.md
```diff
@@ -0,0 +1,34 @@
...
+
+    a. Setup access to the cluster
+
+    b. Get logs and filter by the prefix of the message
+
+        ```sh
+        kubectl -n codacy logs deployment/codacy-api | grep -i "Your database is using key"
+        ```
+        ```
+2. Copy the value and update it in your values yaml file
```
afsmeira, line 25: Should we mention the specific config property to change?

docs/configuration/secrets/troubleshoot.md
afsmeira: That is not great @jllopes . Ideally we would have it in a script, but the script would still need to have the options mentioned above as input.

docs/configuration/secrets/troubleshoot.md
afsmeira: Yes, I think it will help.

docs/configuration/secrets/troubleshoot.md
afsmeira, line 15: Three questions:

- Why mention tail specifically? I think "Get the latest logs from the api service" would be more appropriate.
- Not sure if we are using 'api' or 'API'.
- Guarantee that we always use 'api service'  when referring to website's backend.

docs/configuration/secrets/troubleshoot.md
rtfpessoa: ths logs has it

docs/configuration/secrets/troubleshoot.md
rtfpessoa: > Guarantee that we are always calling 'api service' to the website backend

Not sure how to do this.

The only reason I called it api is because the name of the pod is api.

rtfpessoa:
@jllopes @afsmeira can you test the script please.
Seems good to me.

docs/configuration/secrets/troubleshoot.md
pauloribeiro-codacy: The `(...options used to install Codacy...)` bit was already meant to solve the issue of the command becoming outdated. Do you think that even the `helm upgrade` is subject to change?

docs/configuration/secrets/extract-codacy-secrets.sh
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env bash
+
+usage()
+{
+    echo "Usage: $0 -n <NAMESPACE>"
+    echo "Example: '$0 -n codacy' would retrieve secrets from the logs from the 'codacy' namespace of the kubernetes cluster"
```
pauloribeiro-codacy, line 6: I'm thinking we could have an extra line just to say what this script does. And then the example could become a bit simpler:

```suggestion
    echo "Retrieves secrets used to encrypt sensitive data on the database from the Codacy logs.
    echo "Usage: $0 -n <NAMESPACE>"
    echo "Example: '$0 -n codacy' retrieves secrets from Codacy running on the cluster namespace 'codacy'"
```

Does this make sense?

docs/configuration/secrets/extract-codacy-secrets.sh
```diff
@@ -0,0 +1,65 @@
...
+
+echo "Starting secrets extraction"
+
+# Check if kubectl or microk8s.kubectl is available
+echo "Checking if kubectl is installed..."
+KUBECTL=$(which kubectl || which microk8s.kubectl)
+if [ $? -ne 0 ]; then
+    echo "kubectl not installed"
+    echo "Please install kubectl version specified in Codacy's documentation (see here - https://codacy.github.io/chart/install/) or the version used when installing your cluster"
+    echo "To install kubectl see https://kubernetes.io/docs/tasks/tools/install-kubectl/ (or https://microk8s.io/docs/ if you are running a microk8s kubernetes cluster)"
```
pauloribeiro-codacy, line 31: Small edits to simplify the copy a bit:

```suggestion
    echo "Please install the kubectl version specified in Codacy's documentation (https://codacy.github.io/chart/install/) or the version used when installing your cluster"
    echo "To install kubectl see https://kubernetes.io/docs/tasks/tools/install-kubectl/, or https://microk8s.io/docs/ if you are running a MicroK8s kubernetes cluster"
```

docs/configuration/secrets/extract-codacy-secrets.sh
```diff
@@ -0,0 +1,65 @@
...
+echo "Checking access to kubernetes cluster..."
+KUBE_CTX=$($KUBECTL config current-context)
+if [ $? -ne 0 ]; then
+    echo "No kubernetes cluster context configured"
+    exit 5
+fi
+
+read -p "Is '$KUBE_CTX' the correct kubernetes cluster for log extraction? (yes/[no]): " ANSWER
+if [[ ! "$ANSWER" =~ ^y(es)?$ ]]; then
+    echo "Please configure correctly your current kubernetes cluster"
```
pauloribeiro-codacy, line 45: Should this message mention something about configuring the cluster context (https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/)?

```suggestion
    echo "Configure your kubernetes cluster context to access the cluster where Codacy is running"
```

docs/configuration/secrets/extract-codacy-secrets.sh
```diff
@@ -0,0 +1,65 @@
...
+    echo "Failed to get the 'codacy-api' deployment, for namespace $NAMESPACE"
+    echo "Are you sure you are in the right kubernetes cluster context?"
+    exit 7
+fi
+
+DEPLOYMENT_LOGS_COMMAND="$KUBECTL -n $NAMESPACE logs deployment/codacy-api"
+SECRET_MESSAGES=$($DEPLOYMENT_LOGS_COMMAND | grep -i "Your database is using key")
+if [ $? -ne 0 ]; then
+    echo "Failed to get the logs for the 'codacy-api' deployment, for namespace $NAMESPACE"
+    echo "You can try running the command by hand ('$DEPLOYMENT_LOGS_COMMAND') or contacting support@codacy.com"
```
pauloribeiro-codacy, line 61: Running the command manually probably won't make much of a difference, right? :disappointed_relieved:

docs/configuration/secrets/troubleshoot.md
```diff
@@ -0,0 +1,35 @@
+# Troubleshooting
+
+The sections below include information to help you troubleshoot issues that you may come across while configuring the secrets used by Codacy.
+
+If the information provided on this page is not enough to solve your issue, contact [support@codacy.com](mailto:support@codacy.com) providing all the information that you were able to collect while following the troubleshooting instructions.
```
pauloribeiro-codacy, line 5: This is a more generic comment, but I'm thinking we should check with Support if they're comfortable whenever we suggest to the users that they should use support@codacy.com to contact Codacy. At the very least, Support should know about these "entry points" so they are empowered to tackle the tickets from customers more efficiently. (@IGFCoimbra, @paulacodacy).

docs/configuration/secrets/troubleshoot.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Lost/Changed initially generated secrets
+
+While trying to open Codacy,
+you were shown an error message stating that,
+the secret on the database and the one in your configuration file are different.
```
pauloribeiro-codacy, line 11:
```suggestion
When you open the Codacy UI, an error message states that the secret on the database and the one in your configuration file are different.
```

docs/configuration/secrets/troubleshoot.md
```diff
@@ -0,0 +1,35 @@
...
+
+### Steps
+
+1. Get the latest logs from the API service
+
+    a. Get logs and filter by the prefix of the message
```
pauloribeiro-codacy, line 17: I would remove this detail (since the script does a great job in abstracting this) and unindent the lines below one level.

docs/configuration/secrets/extract-codacy-secrets.sh
```diff
@@ -0,0 +1,65 @@
...
+echo "Checking if the deployment exists..."
+DEPLOYMENT_INFORMATION=$($KUBECTL -n $NAMESPACE get deployments | grep codacy-api)
+if [ $? -ne 0 ]; then
+    echo "Failed to get the 'codacy-api' deployment, for namespace $NAMESPACE"
+    echo "Are you sure you are in the right kubernetes cluster context?"
+    exit 7
+fi
+
+DEPLOYMENT_LOGS_COMMAND="$KUBECTL -n $NAMESPACE logs deployment/codacy-api"
+SECRET_MESSAGES=$($DEPLOYMENT_LOGS_COMMAND | grep -i "Your database is using key")
```
pauloribeiro-codacy, line 58: Don't forget to change the pattern here if you do change the copy when logging the message.

docs/configuration/secrets/troubleshoot.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Lost/Changed initially generated secrets
+
+While trying to open Codacy,
+you were shown an error message stating that,
+the secret on the database and the one in your configuration file are different.
+
+### Steps
+
+1. Get the latest logs from the API service
```
pauloribeiro-codacy, line 15: Since the script automatically extracts the value from the logs, we can simplify this a bit:

```suggestion
1. Obtain the correct key used to encrypt sensitive data in the database from the Codacy logs.
```

docs/configuration/secrets/troubleshoot.md
```diff
@@ -0,0 +1,35 @@
...
+    a. Get logs and filter by the prefix of the message
+
+        ```bash
+        bash <(curl -fsSL https://raw.githubusercontent.com/codacy/chart/master/docs/configuration/secrets/extract-codacy-secrets.sh) \
+            -n <namespace>
+        ```
+
+        You can also download the script [extract-codacy-secrets.sh](extract-codacy-secrets.sh) to run it manually.
+
+2. Copy the value and update it in your values yaml file
```
pauloribeiro-codacy, line 26: We can be just a bit more specific:

```suggestion
2. Copy the value of the key and update your `values-production.yaml` file with this value.
```

docs/configuration/secrets/troubleshoot.md
```diff
@@ -0,0 +1,35 @@
...
+
+3. Restart the application with the new configuration
```
pauloribeiro-codacy, line 28: For consistency we can use exactly the same text here as we are using in the multiple existing places in the documentation (see step 4 of [this page](https://codacy.github.io/chart/configuration/integrations/github-cloud/), for instance).

If later we decide to extract and centralize this step somewhere else, it makes it easier to do the refactor if we already know that the instruction is exactly the same everywhere.

mkdocs.yml
```diff
@@ -55,6 +55,8 @@ nav:
       - configuration/integrations/bitbucket-server.md
       - configuration/integrations/email.md
       - configuration/integrations/troubleshoot.md
+    - "Secrets":
+      - configuration/secrets/troubleshoot.md
```
pauloribeiro-codacy, line 59: I will like to place this new troubleshooting information in a better place (just so we don't have several scattered "troubleshooting" pages). But I'll wait until we settle all the troubleshooting info that we will provide to propose a new / simpler structure.

docs/configuration/secrets/extract-codacy-secrets.sh
rtfpessoa: Yup, that is one of the parts I do not like.
This being tied to the other.
I will put a link on the other side pointing here.

docs/configuration/secrets/extract-codacy-secrets.sh
rtfpessoa: Probably not, but might give you insight into other problems that you can then follow through normal troubleshooting.

docs/configuration/secrets/troubleshoot.md
rtfpessoa: As we prefer, I have no strong opinion about this.
I was just putting this were it makes sense for our customers to know where to go for help.

docs/configuration/secrets/troubleshoot.md
rtfpessoa: I do not think it will frequently.

mkdocs.yml
pauloribeiro-codacy: I opened the pull request https://github.com/codacy/chart/pull/391 to do this refactoring.

PR #397, opened 2020-05-06
fix: Cleanup helm extra flags

h314to:
Cleaning up the HELM_EXTRA_FLAGS field. The variable names are very systematic, so we can use a script to get the specific environment variable form the environment name and the variable root name.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,6 +47,15 @@ references:
 
   ##################################################################################
   #  Environment variables that differentiate for each environment/installation
+  #
+  #  To add environment specific helm values:
+  #    1. Add the prototype to `.do/values-env-temp.yaml`
+  #       Here this must be the common, root name (say, SOMEVAR)
+  #
+  #    2. Create the variables in this project's settings on CircleCI.
+  #       A suffix with the environment name must be appended to the variable name
+  #       (say, SOMEVAR_DEV, SOMEVAR_SANDBOX, SOMEVAR_RELEASE)
+  #
   ##################################################################################
   dev_environment: &dev_environment
     DOKS_CLUSTER_NAME: codacy-doks-cluster-dev
@@ -56,7 +65,8 @@ references:
     ENVIRONMENT: dev
     HELM_REPOSITORY: codacy-unstable
     CODACY_URL: http://dev.k8s.dev.codacy.org
-    HELM_INSTALL_EXTRA_FLAGS: --set global.github.app.name=codacy-k8s-selfhosted-dev --set global.github.app.id="${GITHUB_APP_ID_DEV}" --set global.github.app.privateKey="${GITHUB_PRIVATE_KEY_DEV}" --set global.github.clientId=${GITHUB_CLIENTID_DEV} --set global.github.clientSecret=${GITHUB_CLIENTSECRET_DEV} --set global.githubEnterprise.app.name=codacy-k8s-selfhosted-dev --set global.githubEnterprise.app.id="${GITHUB_ENTERPRISE_APP_ID_DEV}" --set global.githubEnterprise.app.privateKey="${GITHUB_ENTERPRISE_PRIVATE_KEY_DEV}" --set global.githubEnterprise.clientId=${GITHUB_ENTERPRISE_CLIENTID_DEV} --set global.githubEnterprise.clientSecret=${GITHUB_ENTERPRISE_CLIENTSECRET_DEV} --set global.bitbucket.key="${BITBUCKET_KEY_DEV}" --set global.bitbucket.secret="${BITBUCKET_SECRET_DEV}" --set global.bitbucketEnterprise.consumerKey="${BITBUCKET_ENTERPRISE_CONSUMER_KEY_DEV}" --set global.bitbucketEnterprise.consumerPublicKey="${BITBUCKET_ENTERPRISE_CONSUMER_PUBLIC_KEY_DEV}" --set global.bitbucketEnterprise.consumerPrivateKey="${BITBUCKET_ENTERPRISE_CONSUMER_PRIVATE_KEY_DEV}" --set global.gitlab.clientId="${GITLAB_CLIENTID_DEV}" --set global.gitlab.clientSecret="${GITLAB_CLIENTSECRET_DEV}" --set global.gitlabEnterprise.clientId="${GITLABENTERPRISE_CLIENTID_DEV}" --set global.gitlabEnterprise.clientSecret="${GITLABENTERPRISE_CLIENTSECRET_DEV}"
+    GITHUB_APP_NAME_DEV: "codacy-k8s-selfhosted-dev"
+    GITHUB_ENTERPRISE_APP_NAME_DEV: "codacy-k8s-selfhosted-dev"
 
   sandbox_environment: &sandbox_environment
     DOKS_CLUSTER_NAME: codacy-doks-cluster-dev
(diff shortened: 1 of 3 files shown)


.do/values-env-temp.yaml
```diff
@@ -0,0 +1,34 @@
...
+    consumerKey: "${BITBUCKET_ENTERPRISE_CONSUMER_KEY}"
+    consumerPublicKey: "${BITBUCKET_ENTERPRISE_CONSUMER_PUBLIC_KEY}"
+    consumerPrivateKey: "${BITBUCKET_ENTERPRISE_CONSUMER_PRIVATE_KEY}"
+
+  gitlab:
+    clientId: "${GITLAB_CLIENTID}"
+    clientSecret: "${GITLAB_CLIENTSECRET}"
+
+  gitlabEnterprise:
+    clientId: "${GITLABENTERPRISE_CLIENTID}"
```
peavers, line 33: Split the word GITLABENTERPRISE -> GITLAB_ENTERPRISE

.do/values-env-temp.yaml
```diff
@@ -0,0 +1,34 @@
...
+global:
+  github:
+    app:
+      name: ${GITHUB_APP_NAME}
+      id: "${GITHUB_APP_ID}"
+      privateKey: "${GITHUB_PRIVATE_KEY}"
+    clientId: ${GITHUB_CLIENTID}
+    clientSecret: ${GITHUB_CLIENTSECRET}
+
+  githubEnterprise:
```
peavers, line 11: What is enterprise in this context?

.circleci/config.yml
```diff
@@ -171,17 +183,29 @@ references:
...
+          command: make -e -C .do/ deploy_to_doks_from_local_dir HELM_INSTALL_EXTRA_FLAGS="-f values-env.yaml"
 
   deploy_to_cluster_from_chartmuseum: &deploy_to_cluster_from_chartmuseum
     steps:
       - <<: *attach_workspace
       - <<: *doctl_authenticate
+      - run:
+          name: Set environment specific helm values
+          command: |
+            # DELETE ME: this is a temporary fix because breaking changes were added to ci-do so we can't add gettext
```
peavers, line 203: Delete what exactly, the entire command block? The entire run block?

.do/values-env-temp.yaml
```diff
@@ -0,0 +1,34 @@
...
+      id: "${GITHUB_ENTERPRISE_APP_ID}"
+      privateKey: "${GITHUB_ENTERPRISE_PRIVATE_KEY}"
+    clientId: ${GITHUB_ENTERPRISE_CLIENTID}
+    clientSecret: ${GITHUB_ENTERPRISE_CLIENTSECRET}
+
+  bitbucket:
+    key: "${BITBUCKET_KEY}"
+    secret: "${BITBUCKET_SECRET}"
+
+  bitbucketEnterprise:
```
peavers, line 23: Is this Bitbucket Server?

.do/values-env-temp.yaml
```diff
@@ -0,0 +1,34 @@
+# environment specific values
+global:
+  github:
+    app:
+      name: ${GITHUB_APP_NAME}
+      id: "${GITHUB_APP_ID}"
+      privateKey: "${GITHUB_PRIVATE_KEY}"
+    clientId: ${GITHUB_CLIENTID}
+    clientSecret: ${GITHUB_CLIENTSECRET}
```
peavers, line 9: Can we split the word CLIENTSECRET to CLIENT_SECRET? Likewise with CLIENTID -> CLIENT_ID. This applies throughout the file, there are a few cases of it not being consistent :)

.circleci/config.yml
```diff
@@ -171,17 +183,29 @@ references:
             echo "$(date +%s)-$(cat .version)-dev-$(git rev-parse --short HEAD)" > .version
             cat .version
             ytool -s appVersion "$(cat .version)" -s version "$(cat .version)" -f "./codacy/Chart.yaml" -e
+      - run:
+          name: Set environment specific helm values
+          command: |
+            # DELETE ME: this is a temporary fix because breaking changes were added to ci-do so we can't add gettext
```
peavers, line 189: Delete what exactly, the entire command block? The entire run block?

.circleci/config.yml
```diff
@@ -75,7 +86,8 @@ references:
...
+    GITHUB_APP_NAME_RELEASE: "codacy-k8s-selfhost-release"
+    GITHUB_ENTERPRISE_APP_NAME_RELEASE: "codacy-k8s-selfhost-release"
```
peavers, line 90: Should this be codacy-k8s-selfhost**ed**-release?

.circleci/setupEnvironmentVars.sh
```diff
@@ -0,0 +1,21 @@
...
+set -euo pipefail
+
+INFILE=$1
+
+# get lines with a $, split by : and get the second field (i.e. where yaml value, the env variable name, is)
+# and clean it up by stripping whitespace, {, }, and "
+VARS=$(grep '\$' $INFILE | cut -d: -f2 | sed -e 's|[\$\{\}\"\s]||g')
+
+# uppercase the environment name and use it as suffix
+SUFFIX=${ENVIRONMENT^^}
```
joaocosta-codacy, line 12: Just out of curiosity, what does the ^^ do? :)

.circleci/config.yml
joaocosta-codacy, line 203: When could we delete this?

.do/values-env-temp.yaml
h314to: I'm reusing the existing variables and names (which were already added to CircleCI). I'm just cleaning up how they are passed to helm.

.circleci/config.yml
h314to: Nope, just the installation of the package. I can't add a package to ci-do without losing helm2, which I need.

.circleci/setupEnvironmentVars.sh
h314to: It uppercases the variable

.circleci/config.yml
h314to: Yes, it is inconsistent, but if you take a look at the diff that's what's being used for those github apps. Can't change without braking stuff.

.circleci/config.yml
h314to: When we have everything with helm3 so that the package can be added to ci-do, or if we decide to do a more piecewise migration and have helm2 and helm3 side by side for a while.

.circleci/config.yml
h314to: As above, just delete the installation of the package. I can't add a package to ci-do without losing helm2, which I need.

.do/values-env-temp.yaml
h314to: Github self-hosted, which is mapped to the codacy-api environment in the configmap.

.do/values-env-temp.yaml
h314to: Guess so

.do/values-env-temp.yaml
h314to: I'm just refactoring how these variables are passed to helm, using a values file and some standards form variable naming, instead of having repeated code and huge lines with all variables mapping. Maybe the actual naming of the variables can be changed in a subsequent step.

.circleci/config.yml
rtfpessoa: Isn't this the bug machado found on friday https://github.com/codacy/chart/pull/436 ?

h314to: Thanks for the heads up! 👍

h314to:
Things seem to be working well as they are, so there isn't any big driving force to change it now. Plus, this is really outdated and would probably need major revisions. I'm closing this for now. I'll take another look at this if the need arises.

PR #399, opened 2020-05-07
clean: Change configuration properties to have the same order as the installation guide [CY-1582]

afsmeira:


codacy/values-production.yaml
```diff
@@ -363,3 +292,69 @@ minio:
   persistence:
     enabled: true
     size: 20Gi
+
+github: {}
```
pauloribeiro-codacy, line 296: This block is missing some indentation I believe:

```suggestion
  github: {}
```

afsmeira: It was properly indented, but the following ones were not. 👍

PR #412, opened 2020-05-11
fix: Evaluate worker version after helm dep up

joaocosta-codacy:


joaocosta-codacy:
With this, we resolve the engine version only after having helm dep up executed successfully.

PR #422, opened 2020-05-12
Fix: Revert documentation to Helm v2 until v3 is stable

h314to:


machadoit:
I think this is a step backwards, and that we should link the version of the client needs to use. But we could make it work with either approach

h314to:
Clients are installing as we speak and having problems because this is not in sync with the stable version. Right now we don't have versioning for our documentation so reverting is the step back we need to take.

PR #433, opened 2020-05-15
fix: Disable Merge messages on changelogs.

joaocosta-codacy:
This shall only get merged after the current ongoing release.

### DIFF
--- a/.changelogs/gitchangelogtemplate.rc
+++ b/.changelogs/gitchangelogtemplate.rc
@@ -201,7 +201,7 @@ output_engine = mustache("##PATHMACRO##")
 ##
 ## This option tells git-log whether to include merge commits in the log.
 ## The default is to include them.
-include_merge = True
+include_merge = False
 
 
 ## ``log_encoding`` is a string identifier


.changelogs/gitchangelogtemplate.rc
```diff
@@ -201,7 +201,7 @@ output_engine = mustache("##PATHMACRO##")
 ##
 ## This option tells git-log whether to include merge commits in the log.
 ## The default is to include them.
-include_merge = True
+include_merge = False
```
pauloribeiro-codacy, line 204: Here we are recommending allowing merge commits:

https://handbook.dev.codacy.org/product/engineering/guides/github-migration.html#configure-the-merge-button

If we stop including these merge commits, we will lose the information on commits coming from branches, and the changelogs will not be complete. I just tested this with the [Handbook repository](https://github.com/codacy/handbook), for example, and with the setting set to `False` there is a lot of information that is left out. :thinking:

pauloribeiro-codacy: OK, in the meanwhile @joaocosta-codacy mentioned an important bit of information that I was missing, which is the fact that the `gitchangelog` tool fetches the commit messages from all the branches and not just from `master`, and this is why the commit messages are duplicated in the changelogs. :+1:

PR #436, opened 2020-05-15
fix: Create new codacy-k8s-selfhosted-release Github App and add missing quotes

machadoit:
I've added some screenshots using a very nice tool, because it is hard to see the diff :D

image

image


rtfpessoa:
Why doesn't it fail with helm during deploy with missing key?

machadoit:
@rtfpessoa keep in mind that I don't know if this was really necessary! It was to make sure it was equal to the others. The point here, is that this uses a newly github application

PR #437, opened 2020-05-15
fix-integration-identation

heliocodacy:


codacy/values-production.yaml
```diff
@@ -65,6 +66,72 @@ global:
     service:
       port: <--- codacy-db-port --->
 
+  github: {}
```
rtfpessoa, line 69: @afsmeira didn't you change this order as requested by @pauloribeiro-codacy ?

rtfpessoa: Why are you changing this @heliocodacy ?

gjsduarte: The indentation as-is places the provider configurations wrongly outside the global section. 
As we can also not repeat the global section, the providers must be placed in this location of the file.

PR #455, opened 2020-05-21
fix: Revert servicemonitor creation due to prometheus operator helm3 issues

h314to:
With helm3 if service monitors are enabled deployment fails with the following error:

```
Error: UPGRADE FAILED: error validating "": error validating data: [ValidationError(ServiceMonitor.spec.selector): unknown field "app.kubernetes.io/instance" in com.coreos.monitoring.v1.ServiceMonitor.spec.selector, ValidationError(ServiceMonitor.spec.selector): unknown field "app.kubernetes.io/name" in com.coreos.monitoring.v1.ServiceMonitor.spec.selector]
  
```

I'm reverting this for now until we find a workaround and fix the components' servicemonitor templates.

### DIFF
--- a/docs/configuration/monitoring.md
+++ b/docs/configuration/monitoring.md
@@ -1,11 +1,15 @@
 # Monitoring
 
+<!---
+FIXME: Commented out to prevent issues while service monitors are not compatible with helm3
+
 Currently, we support two monitoring solutions:
 
 -   **[Crow](#setting-up-monitoring-using-crow):** A simple, lightweight, and built-in monitoring solution, that is enabled by default when you install Codacy.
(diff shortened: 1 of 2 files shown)


paulopontesm:
This means that the section for monitoring doesn't work, right?

docs/configuration/monitoring.md
```diff
@@ -118,13 +118,13 @@ Now that you have Prometheus and Grafana installed you can enable `serviceMonito
     codacy-api:
       metrics:
         serviceMonitor:
-          enabled: true
```
paulopontesm, line 121: This dosent match the scentence above:

Now that you have Prometheus and Grafana installed you can enable `serviceMonitor`.

Maybe we should remove this for now

h314to: agreed 👍

h314to: I commented out the grafana documentation while our service monitors are not compatible with helm 3.

PR #456, opened 2020-05-22
Remove helm setup from eks quickstart

paulopontesm:


paulopontesm:
> This is a really nice simplification to the procedure required to set up Codacy on EKS! 🙌

And since we moved to helm 3 this didn't make sense anymore because it was installing helm 2.

docs/infrastructure/EKS/backend/variables.tf
```diff
@@ -2,12 +2,12 @@
...
+  type        = string
+  default     = "codacy"
 }
 
 variable "custom_tags" {
   description = "Map of custom tags to apply to every resource"
-  type = map(string)
-  default = null
+  type        = map(string)
+  default     = {}
```
h314to, line 12: I think this was working well with `null` (cf. https://www.terraform.io/docs/configuration/expressions.html#null ). Did you run into some issues?

docs/infrastructure/EKS/main/cluster_masters.tf
```diff
@@ -30,11 +30,46 @@ resource "aws_eks_cluster" "main" {
 
 }
 
+# Get an authentication token to communicate with an EKS cluster, using the "kubernetes" provider.
+data "aws_eks_cluster_auth" "eks_cluster_auth" {
+  name = "${aws_eks_cluster.main.name}"
+}
+
+provider "kubernetes" {
```
h314to, line 38: Please move provider block to the `config.tf` file. This way it is easier to keep track of where everything is coming from.

docs/infrastructure/EKS/main/cluster_masters.tf
```diff
@@ -30,11 +30,46 @@ resource "aws_eks_cluster" "main" {
 
 }
 
+# Get an authentication token to communicate with an EKS cluster, using the "kubernetes" provider.
+data "aws_eks_cluster_auth" "eks_cluster_auth" {
```
h314to, line 34: Please move data resources to the `constants.tf` file. This might lead to a cyclic dependency. This might not be needed. See below.

docs/infrastructure/EKS/main/cluster_masters.tf
```diff
@@ -30,11 +30,46 @@ resource "aws_eks_cluster" "main" {
...
+  name = "${aws_eks_cluster.main.name}"
+}
+
+provider "kubernetes" {
+  host                   = "${aws_eks_cluster.main.endpoint}"
+  cluster_ca_certificate = "${base64decode(aws_eks_cluster.main.certificate_authority.0.data)}"
+  token                  = "${data.aws_eks_cluster_auth.eks_cluster_auth.token}"
+  load_config_file       = false
+  version                = "~> 1.5"
+}
```
h314to, line 44: I'm not sure this would work. The cluster configuration only exists after the cluster was created, but providers are setup when you run `terraform init` (i.e., before the cluster is created). The first run would probably fail, and you'd have to run it second time for it to work.

I think it would be better to provide the correct configmap as an output and run kubectl apply on it, thus avoiding the bootstrap issue. This already exists. Just run:
```
terraform output aws_auth_configmap | kubectl apply -f -
```
after everything is created.

docs/infrastructure/EKS/main/variables.tf
```diff
@@ -3,138 +3,138 @@
...
-  type = string
-  default = "codacy"
+  type        = string
+  default     = "codacy"
 }
 
 variable "custom_tags" {
   description = "Map of custom tags to apply to every resource"
-  type = map(string)
-  default = null
```
h314to, line 19: As above, was null not working?

docs/infrastructure/EKS/backend/variables.tf
paulopontesm: Yep, got an error saying it cant be null :/

docs/infrastructure/EKS/main/cluster_masters.tf
paulopontesm: I tried, it worked

paulopontesm: In fact I copied from the internets.. some guy in github did this in his template. It made sense to me and seems to work.

docs/infrastructure/EKS/backend/variables.tf
paulopontesm: These is the error:
image

Terraform versions:
image

h314to: 👍 I figured you wouldn't be fixing it if it wasn't broken, and was curious.

docs/infrastructure/EKS/main/cluster_masters.tf
h314to: If it works, great. Maybe it was something was fixed recently. I seem to recall this type of thing being an issue. Make sure it works on a clean installation, and remove the `.terraform/` to test, if it exists.

docs/infrastructure/eks-quickstart.md
```diff
@@ -97,53 +96,12 @@ Create a cluster that includes all the required network and node setup:
 
     You'll notice that nothing is scheduled. That's because we haven't yet allowed the worker nodes to join the cluster (see the [EKS docs](https://docs.aws.amazon.com/eks/latest/userguide/add-user-role.html) for more details). We'll do that on the next section.
```
h314to, line 97: This is no longer valid.

docs/infrastructure/EKS/main/cluster_masters.tf
paulopontesm: Done that. It seems like it works.

PR #463, opened 2020-05-25
feature: Remove AKS docs and infrastructure

h314to:


h314to:
We would now have to refactor this now anyway due to shifting support from helm2 to helm3. This is unmaintained, not deployed anywhere, and unused. It is, therefore, hard to keep it up to par with the EKS or Microk8s infrastructure in terms of quality and overall stability. We're taking the cautionary approach of removing it for now. If in the future we need to support AKS we can use these stacks as a starting point, but we would really have to deploy and test frequently in AKS to ensure it is production ready.

pauloribeiro-codacy:
> We would now have to refactor this now anyway due to shifting support from helm2 to helm3. This is unmaintained, not deployed anywhere, and unused. It is, therefore, hard to keep it up to par with the EKS or Microk8s infrastructure in terms of quality and overall stability. We're taking the cautionary approach of removing it for now. If in the future we need to support AKS we can use these stacks as a starting point, but we would really have to deploy and test frequently in AKS to ensure it is production ready.

Makes total sense. Thanks, @h314to! :+1:

PR #464, opened 2020-05-25
fix: aws_awth for eks

paulopontesm:


paulopontesm:
image

This is the relevant line. As it is, the nodes cant add themselves on bootstrap

PR #470, opened 2020-05-29
Docs versioning

paulopontesm:
This changes the CI to publish the docs only when a release is done.

This is how it looks:

image

image


rtfpessoa:
We could probably publish them to circleci artifacts everytime, so we could see it rendered there

.circleci/config.yml
```diff
@@ -361,7 +361,7 @@ jobs:
           name: Set cluster
           command: |
             sudo pip install -r requirements.pip
-            mkdocs gh-deploy -m "[ci skip] Updated documentation."
+            mike deploy -u $(cat .version) latest -m "[ci skip] Updated documentation."
```
rtfpessoa, line 364: What is this mike?

rtfpessoa: https://github.com/jimporter/mike

docs/js/version-select.js
```diff
@@ -0,0 +1,63 @@
+window.addEventListener("DOMContentLoaded", function () {
+  // This is a bit hacky. Figure out the base URL from a known CSS file the
+  // template refers to...
+  var ex = new RegExp("/?assets/fonts/material-icons.css$");
```
rtfpessoa, line 4: Where does `material-icons` como from? Cannot find it mentioned anywhere

paulopontesm:
part of the css/javascript is generated by the "mike" tool, part is to make the select box more "material like"

.circleci/config.yml
paulopontesm: yes, that mike

paulopontesm: also added to the requirements.pip

docs/js/version-select.js
paulopontesm: this is generated when I run the `mike install-extras` 

https://github.com/jimporter/mike#building-your-docs

paulopontesm: I still edited this file to add the classes to do the <select> more "material like" .. by default it creates the ugly default <select>

.circleci/config.yml
pauloribeiro-codacy: This basically allows you to deploy several versions of the documentation in parallel to the `gh-pages` branch (while `mkdocs`, on the other hand, always overwrites the files in the branch). :+1:

docs/js/version-select.js
pauloribeiro-codacy: Is this working on this repository? When I tried this on the codacy/docs repository, it couldn't find the `material-icons.css` file and the whole JavaScript would fail.

I completely rewrote this `version-select.js` file to make it way less hacky, but I still need to solve the problem of the selected version not being kept between page reloads:

https://github.com/codacy/docs/pull/31/files

pauloribeiro-codacy:
I'm trying to solve the same problem for the new codacy/docs repository here:

https://github.com/codacy/docs/pull/31/

This repository includes the codacy/chart as a submodule and once we get it up and running it will centralize the user documentation and the installation documentation in the same site. But while this doens't happen (it could take a few more weeks), I say that we start versioning the docs here in the codacy/chart and then apply what we've learned to the codacy/docs. :+1:

PR #471, opened 2020-05-29
Feature/add ingress docs

h314to:


codacy/values-tls-ingress-aws.yaml
```diff
@@ -0,0 +1,26 @@
+## For guidance on how to setup tls please refer to the documentation on https://codacy.github.io/chart
```
pauloribeiro-codacy, line 1:
```suggestion
## For guidance on how to set up TLS please refer to the documentation on https://codacy.github.io/chart
```

docs/index.md
```diff
@@ -144,3 +144,4 @@ After successfully installing Codacy on your cluster, you are now ready to perfo
 3.  As a last step we recommend that you [set up monitoring](configuration/monitoring.md) on your Codacy instance.
 
 If you run into any issues while configuring Codacy, be sure to [check our troubleshooting guide](troubleshoot/troubleshoot.md) for more help.
+If you with to setup HTTPS ingress for Codacy on AWS please see see our [installation guide](configuration/tls-ingress-aws.md).
```
pauloribeiro-codacy, line 147:
```suggestion
3.  If you are using Amazon EKS as your Kubernetes infrastructure and you wish to set up HTTPS ingress for Codacy, follow our [configuration instructions](configuration/tls-ingress-aws.md).

4.  As a last step we recommend that you [set up monitoring](configuration/monitoring.md) on your Codacy instance.

If you run into any issues while configuring Codacy, be sure to [check our troubleshooting guide](troubleshoot/troubleshoot.md) for more help.
```

pauloribeiro-codacy: What do you think of being a bit more prescriptive and including this configuration as part of the "post-install configuration steps", even though it's up to the customer to decide if they want to have HTTPS or not?

pauloribeiro-codacy: Can we also make this work with MicroK8s on Amazon EKS? We would need to instruct using `microk8s.kubectl` instead of `kubectl` on the appropriate commands.

pauloribeiro-codacy:
@h314to, this is an awesome guide on how to configure HTTPS, since the configuration that must be performed manually by the user is so simple. :raised_hands: 

I edited the instructions to make them even more streamlined and also importantly, to give a bit of context about the configuration right at the top of the page.

Please review and let me know if you don't agree with anything or if there is something else we could improve.

docs/index.md
h314to: Initially I left it out of the basic setup because it has a few moving parts, it isn't something everyone will need, and users might want to setup their certificates differently. But your suggested edit seems fine because it conveys the message that this is not a requirement.

Regarding microk8s support, we should do it, but I think it is better to split this into two tasks to unblock users on AWS that currently require this. Afterwards we can broaden the scope of the docs to encompass microk8s as well. I don't think it will require much work, I just haven't tested this sufficiently there. WDYT?

pauloribeiro-codacy: It makes total sense, let's release this now :rocket: and work on the configuration MicroK8s later. I'm hopeful that the configuration for MicroK8s won't be too different from this one.

pauloribeiro-codacy:
@h314to , the `helm_lint` test is failing but I can't understand why. Perhaps it's something simple to fix:

```
==> Linting codacy/
[INFO] Chart.yaml: icon is recommended
[ERROR] templates/: chart metadata (Chart.yaml) missing

Error: 1 chart(s) linted, 1 chart(s) failed

Exited with code exit status 1
```

h314to:
Yes, it is caused by the negated file in `.helmignore`. This is [supposedly supported](https://helm.sh/docs/chart_template_guide/helm_ignore_file/) but is causing a weird error. The error message isn't that helpful... I'll just ignore the files one by one.

This prevents their inclusion in the helm package, but maybe in the future we should should consider using a different folder for addon values files, to reduce bloat in the chart dir.

docs/index.md
```diff
@@ -141,6 +141,8 @@ After successfully installing Codacy on your cluster, you are now ready to perfo
     -   Creating an initial organization
     -   Inviting users to Codacy
 
-3.  As a last step we recommend that you [set up monitoring](configuration/monitoring.md) on your Codacy instance.
+3.  If you are using Amazon EKS as your Kubernetes infrastructure and you wish to set up HTTPS for Codacy, follow our [configuration instructions](configuration/tls-ingress-aws.md).
```
paulopontesm, line 144: Why just for EKS?

codacy/values-tls-ingress-aws.yaml
```diff
@@ -0,0 +1,26 @@
+## For guidance on how to set up TLS please refer to the documentation on https://codacy.github.io/chart
+
+codacy-ingress:
```
paulopontesm, line 3: This is in the values-production .. do we need all these values files?

docs/index.md
paulopontesm: How do I do it for microk8s?

h314to: Please see my comment [here](https://github.com/codacy/chart/pull/471#discussion_r433247597). Its from a previous commit, so it probably collapsed

codacy/values-tls-ingress-aws.yaml
h314to: This overrides the defaults in values production. The annotations are different and this has more stuff in it. Using a different file simplifies values-production by excluding optional configuration.

docs/index.md
pauloribeiro-codacy: Yes, ideally we should have something also for MicroK8s. As soon as you have time to test the solution you have in mind @h314to, I'm glad to help. :+1:

codacy/values-tls-ingress-aws.yaml
pauloribeiro-codacy: Yes, otherwise the users will (most probably) configure all the values at once in step 4 of this section of the installation instructions:

https://codacy.github.io/chart/latest/#2-installing-codacy

(But I do think that we're not being 100% consistent here, for example, the configurations for the Git providers are all in the `values-production.yaml` even though their configuration is optional since only one is required to be enabled.)

h314to: That's a fair point. We'll need to test and document this for microk8s very shortly, so we can use that chance to try to harmonize the configuration.

PR #473, opened 2020-06-02
nighly support helm3 only and 1.16-18

paulopontesm:


.circleci/config.yml
```diff
@@ -538,53 +592,56 @@ workflows:
...
-            - install_k8s-1.13_helm-2.16
-            - install_k8s-1.14_helm-2.16
-            - install_k8s-1.15_helm-2.16
-            - install_k8s-1.16_helm-2.16
             - install_k8s-1.13_helm-3.2
             - install_k8s-1.14_helm-3.2
             - install_k8s-1.15_helm-3.2
             - install_k8s-1.16_helm-3.2
+            - install_k8s-1.17_helm-3.2
+            - install_k8s-1.18_helm-3.2
```
joaocosta-codacy, line 644: This is really nice! The fact that we can automatically test that we can install on all these versions 👍

PR #474, opened 2020-06-02
fix: Add port value for certain services

afsmeira:
Services should really not be opinionated on what port they run on. That should
be configured when deploying on a specific environment. Since the service Crow
no longer assumes in its configurations on what ports some services are running,
we need to specify that here.

See https://bitbucket.org/qamine/crow/pull-requests/163/explicitly-set-sslmode-when-connecting-to

### DIFF
--- a/codacy/values.yaml
+++ b/codacy/values.yaml
@@ -15,6 +15,7 @@ global:
       url: "codacy-engine"
     listener:
       url: "codacy-listener"
+      port: 80
     activities:
       url: "codacy-activities"
     core:
@@ -23,6 +24,7 @@ global:
       url: "codacy-remote-provider-service"
     workerManager:
       url: "codacy-worker-manager"
+      port: 80
 
   github: {}
     # enabled: "false"


paulopontesm:
I feel like we should just "hide" those and leave the defaults that come from the values file in crow.

A good option is alto to put it everywhere, but we need a comment saying that it needs to match the port of the service.

paulopontesm:
> I feel like we should just "hide" those and leave the defaults that come from the values file in crow.
> 
> A good option is alto to put it everywhere, but we need a comment saying that it needs to match the port of the service.

More on this:

The communication between crow and WM/Listener currently (by default) is internal. It connects directly to the service port.

If we wanted to change this, we also needed to change the `worker-manager.service.port` to match the same value: https://bitbucket.org/qamine/worker-manager/src/414ef0822852611c6d369e15b497dc8a30ec6bb6/.helm/worker-manager/values.yaml#lines-131

Those values could be useful to be configurable if someone starts putting something like the istio ecosystem to have internal communication throught SSL. But we don't have that documented or tested. I am not sure if it could generate problems with codacy-ingress.

PR #475, opened 2020-06-03
fix: Change repository-listener to github

rtfpessoa:


rtfpessoa:
Let's just see if credentials are alright in the other branch before merging this.

PR #484, opened 2020-06-17
fix: Redirect docs root to latest version

h314to:
Currently visiting the root of our docs show an older version of them. With this PR we are setting `latest` as the default docs, which then causes a redirection of the root to `latest` (see [here](https://github.com/jimporter/mike#setting-the-default-version) for more info).  I tested locally and it worked correctly.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -395,6 +395,7 @@ jobs:
           command: |
             sudo pip install -r requirements.pip
             mike deploy -u $(cat .version) latest -m "[ci skip] Updated documentation" -p
+            mike set-default latest
 
   get_changelogs:
     <<: *default_doks_image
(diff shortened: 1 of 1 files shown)


h314to:
I'm also using mike to push the default changes, which I hadn't done in my previous commit.

pauloribeiro-codacy:
I tried this out now but the redirection to latest wasn't working.

This was because an old version of the documentation was deployed on the "root" of the `gh-pages` branch, and it seems that mike does not overwrite the `index.html` file with the redirect script if it already exists.

So, I cleaned up the `gh-pages` branch completely and re-generated the documentation for all versions starting at 1.0.0 until 1.3.0, and set version 1.3.0 with the alias `latest`.

When we open https://codacy.github.io/chart/, a script in the `index.html` automatically redirects the browser to https://codacy.github.io/chart/latest/.

Currently, the version select drop-down does not appear because of a JavaScript error. I will open a new pull request to fix it.

PR #492, opened 2020-06-25
feature: Add codacy-tools

lolgab:


codacy/values-microk8s.yaml
```diff
@@ -53,6 +53,12 @@ engine:
       cpu: 1
       memory: 2Gi
 
+codacy-tools:
+  resources:
+    limits:
+      cpu: 0.5
```
h314to, line 59: Maybe we can reduce this. Current usage is cpu:17m, mem:424MiB, so I would use cpu: 200m, and mem: 512MiB  or something like that. The mem and cpu requirement for microk8s are very tight.

codacy/values-production.yaml
```diff
@@ -295,6 +295,15 @@ engine:
     requests:
       cpu: 200m
       memory: 500Mi
+codacy-tools:
+  replicaCount: 1
+  resources:
```
h314to, line 300: It should be fine to request more resources here, because we can assume a larger cluster is available, so these limits should be ok.

h314to:
```
Error: can't get a valid version for repositories codacy-tools. Try changing the version constraint in Chart.yaml
```

h314to:
in requirements you have, for the version of codacy-tools:
```
version: ">=6.3.0"
```

but the published versions are `>=0.1.0`

h314to:
also check `requirements-dev.yaml`

lolgab:
Thank you for the review 🎉

PR #493, opened 2020-07-02
fix: disable Kamon in codacy-tools by default

h314to:
Kamon is somehow creating an issue while starting up codacy-tools, which makes it crash, and prevents the engine to start. I'm disabling it for now, until it is fixed, so that Codacy deploys cleanly.

### DIFF
--- a/codacy/values.yaml
+++ b/codacy/values.yaml
@@ -318,6 +318,9 @@ codacy-tools:
   fullnameOverride: codacy-tools
   config:
     environment: enterprise
+    metrics:
+      kamon:
+        enabled: "false"
   metrics:
     serviceMonitor:
       enabled: false


andreaTP:
I would like to know more:
 - what's the error that Kamon is throwing in codacy-tools?
 - why engine is not initializing?

aside from this we have a little typo in the commit message: `Kmon -> Kamon` (just to make sure this can be tracked back in case)
I would rather prioritize the fix of codacy-tools instead of merging this if we have a little time-span for it.
WDYT @codacy/toss ?

h314to:
Yes, a full fix would be nice if you find the time for it. Like I commented elsewhere, there isn't a lot of useful data in the logs. Just Kamon starting up, followed by quick termination of codacy-tools. Maybe more data can be gathered increasing de log level. Regarding engine, when it first starts up it tries to connect to codacy-tools. After waiting for a while for it to be ready it gives up and crashes.

If you'd like to replicate the error you can do it by enabling kamon on the codacy-tools configmap, and after that killing codacy-tools and engine pods.

h314to:
Closing this because we are going to disable by default it the codacy-tools chart.

PR #496, opened 2020-07-13
add eks quickstart overview image

paulopontesm:


docs/infrastructure/eks-quickstart.md
```diff
@@ -2,6 +2,11 @@
 
 Follow the instructions below to set up an Amazon EKS cluster from scratch, including all the necessary underlying infrastructure, using Terraform.
 
+![codacy eks quicksart](images/codacy%20chart%20eks%20quickstart.png)
+
+The image above is a non-exhaustive overview of what you can expect to have
+deployed in your AWS account by using this quickstart.
+
```
pauloribeiro-codacy, line 9:
```suggestion
The diagram below presents a non-exhaustive architecture overview of what you can expect to have
deployed on your AWS account by following this quickstart tutorial.

![Codacy EKS quickstart architecture](images/codacy%20chart%20eks%20quickstart.png)
```

pauloribeiro-codacy: It's better to rename the image file so it doesn't include spaces. For instance:

`eks-quickstart-architecture.png`

pauloribeiro-codacy:
@paulopontesm, what tool did you use to generate this diagram? I'm just trying to understand if the "design language" of the diagram is something that customers may already be familiar with.

paulopontesm:
> @paulopontesm, what tool did you use to generate this diagram? I'm just trying to understand if the "design language" of the diagram is something that customers may already be familiar with.

I used draw.io 

https://drive.google.com/file/d/1V0f4Yjdks29y-UYbk6SSAGv1q5tXKuU5/view?usp=sharing (only available w/ codacy email)

It is one of the tools named by AWS and has all their default icons https://aws.amazon.com/architecture/icons/.

You might be noticing some difference for most AWS drawings because they re-designed all the logos in 2019.

docs/infrastructure/eks-quickstart.md
pauloribeiro-codacy: Hi @paulopontesm, I noticed you did a force push to https://github.com/codacy/chart/commit/8b7ae40734b13a269fbeef54975bb35ea6923cce, but the commit is identical to the previous one. Was this intended?

pauloribeiro-codacy:
> I used draw.io: https://drive.google.com/file/d/1V0f4Yjdks29y-UYbk6SSAGv1q5tXKuU5/view?usp=sharing (only available w/ codacy email). It is one of the tools named by AWS and has all their default icons https://aws.amazon.com/architecture/icons/.

Ah, this is nice, because it means that the icons are the official ones. :+1: 

> Also, jf something is wrong or can be improved feel free to do it.

At first glance, I would say that it could be cool if we could simplify the diagram a bit. But I don't have enough knowledge to know what is essential and accessory in the diagram. :sweat_smile: 

What I notice in the current version is that everything seems to be "at the same level of importance", be it individual Codacy components, the network connections between them, and even larger AWS concepts such as VPCs, Availability Zones, Security Groups, etc. This means that the diagram must be rather large to be able to accommodate this level of detail.

We could consider if there is some kind of information that is more relevant to show here - I'm betting it is the AWS EKS architecture provided by following the quickstart guide. We could simplify the diagram a bit to focus just on this "layer" of the architecture, hiding the details from individual Codacy components. The diagram would be easier to read and the image could become smaller so it's easier to fit in the documentation page.

Do you want to give it a go? Or do you consider that all the current detail can be relevant for customers?

pauloribeiro-codacy:
@paulopontesm, I have one last question. Should the connection of the user accessing Codacy go through the Internet and NAT Gateways, and only then through the Load Balancers? Or are the NAT Gateways only used for traffic going out of the Kubernetes nodes? :thinking: 

By the way, I updated the text "Nat Gateway" to "NAT Gateway" [directly on the diagram](https://drive.google.com/file/d/1V0f4Yjdks29y-UYbk6SSAGv1q5tXKuU5/view?usp=sharing).

paulopontesm:
> are the NAT Gateways only used for traffic going out of the Kubernetes nodes

Yes, that's it. The NAT carries all the connections started from the inside. 
You do have traffic both ways, but it is supposed to represent who initiates the connection.

PR #499, opened 2020-07-15
bump: codacy-tools version

andreaTP:


lolgab:
You need to change it in `requirements-dev` too.

PR #502, opened 2020-07-16
feature: Adds gitlab enterprise documentation for system hooks CY-2070

bmbferreira:


pauloribeiro-codacy:
Nice one @bmbferreira, all the relevant information seems to be there. :+1:

PR #506, opened 2020-07-23
Fix codacy tools values to match regex validation

heliocodacy:


codacy/values-microk8s.yaml
```diff
@@ -57,7 +57,10 @@ codacy-tools:
   resources:
     limits:
       cpu: 200m
-      memory: 512MiB
+      memory: 512Mi
```
paulopontesm, line 60: Can you use these as a reference?
https://github.com/codacy/codacy-tools/blob/dev/.helm/codacy-tools/values.yaml#L26

PR #515, opened 2020-07-30
fix: use relative links in docs

h314to:
We're having an issue where the documentation to previous versions is referencing files in the master branch, which may lead to incompatibility with these files as we continue working on them and evolving our installation.

The vast majority of offending files was cluttering (like we've previously discussed) the chart folder, but only really make sense with the context provided by the documentation. As such, I think it is fair to think of them as an integral part of the docs, and move them there. The only exceptions to this are two files, viz. `values-production.yaml` and `values-microk8s.yaml`, which I think should be kept in the chart folder, because of they are more self-documented and easier to understand without the docs. I'd recommend creating a symbolic link of these two files to the new `docs/values-files/` folder.

With these changes the files would be resolved to the correct branch without the need for further configuration or changes to mkdocs.

Let me know what you think.


.do/Makefile
```diff
@@ -123,13 +123,13 @@ deploy_monitoring: set_cluster_context
 	-kubectl create namespace monitoring
```
h314to, line 124: This is the only place where our dev installation files requires changes.

codacy/.helmignore
```diff
@@ -28,14 +28,3 @@
 
 ## addon values files and dev requirements
 requirements-dev.yaml
```
h314to, line 30: We can now clean this up. Notice I'm not excluding `values-microk8s.yaml`. I think, like `values-production.yaml` it makes sense to bundle it in in the chart.

docs/configuration/monitoring.md
```diff
@@ -69,7 +69,7 @@ kubectl apply -f "https://raw.githubusercontent.com/coreos/prometheus-operator/r
 
 ### 2. Installing Loki
 
-Obtain the configuration file for Loki, [`values-loki.yaml`](https://raw.githubusercontent.com/codacy/chart/master/codacy/values-loki.yaml), and install it by running the command below. While the default storage class setting for Loki persistence should suit most use cases, you may need to adjust it to your specific Kubernetes installation. For instance, for MicroK8s use `storageClassName: microk8s-hostpath`.
+Obtain the configuration file for Loki, [`values-loki.yaml`](../values-files/values-loki.yaml), and install it by running the command below. While the default storage class setting for Loki persistence should suit most use cases, you may need to adjust it to your specific Kubernetes installation. For instance, for MicroK8s use `storageClassName: microk8s-hostpath`.
```
h314to, line 72: Using the relative links also makes it look cleaner :)

docs/index.md
```diff
@@ -75,13 +71,7 @@ Install Codacy on an existing cluster using our Helm chart:
 6.  <span id="helm-upgrade">Add Codacy's chart repository to your Helm client and install the Codacy chart using the file `values-production.yaml` created previously.</span>
 
     !!! important
-        **If you are using MicroK8s** you must use the file `values-microk8s.yaml` together with the file `values-production.yaml`.
-
-        Use `wget` to download the extra file and uncomment the last line before running the `helm upgrade` command below:
-
-        ```bash
-        wget https://raw.githubusercontent.com/codacy/chart/master/codacy/values-microk8s.yaml
```
h314to, line 83: I'd remove this command because it forces you to write the path to a static file, or use more complicated parsing just for the sake of teaching someone how to download a file.

docs/values-files/values-microk8s.yaml
```diff
@@ -0,0 +1 @@
+../../codacy/values-microk8s.yaml
```
h314to, line 1: git symbolic link shenanigans

h314to: Git just stores the destination of the symbolic link. It doesn't duplicate the file.

pauloribeiro-codacy:
The move of the `.yaml` files from the folder `codacy` to `docs/values-files` cause the links to these files to become broken on the previously released versions of the documentation.

I have created the issue [DOCS-100](https://codacy.atlassian.net/browse/DOCS-100) to fix this as soon as possible.

PR #522, opened 2020-08-03
refactor: Update values

joaocosta-codacy:


codacy/values.yaml
```diff
@@ -252,8 +252,9 @@ global:
   metrics:
     serviceMonitor:
       enabled: false
-    grafana_dashboards:
+    grafana:
```
paulopontesm, line 255: didn't this change to global?

paulopontesm:
we need to update this path:
	ytool -f "./codacy/values.yaml" -s worker-manager.config.codacy.worker.imageVersion "${ENGINE_VERSION}" -e

codacy/values.yaml
joaocosta-codacy: but this is the global block. I am confused. :)

paulopontesm: sorry .. i'm drunk🤦‍♂️

PR #523, opened 2020-08-03
fix: Change link to repository-listener helm chart to github

h314to:


h314to:
> Why is this the only link pointing to the `dev` branch and not `master`? Is it intended?

Nope, it should be master. Fixed it.

PR #526, opened 2020-08-04
doc: Apply small tweaks based on customer feedback

pauloribeiro-codacy:
This is to address a few comments received from a customer that went through the installation process. In particular:

> I'm following the instructions here: https://codacy.github.io/chart/latest/#2-installing-codacy
> 
> On step 6 it says:
> 
> If you are using MicroK8s you must use the file values-microk8s.yaml together with the file values-production.yaml.
> Use wget to download the extra file and uncomment the last line before running the helm upgrade command below:
> wget https://raw.githubusercontent.com/codacy/chart/master/codacy/values-microk8s.yaml
> 
> However there are no commented lines in values-microk8s.yaml or values-production.yaml. Can I ignore that?
> 
> Also, I need to fill in values for the following, but I don't know where to get them:
> 
> <--- crow-login-password --->
> <--- insert your Codacy license here --->
> <--- generated_secret --->

I'm not addressing the `<--- insert your Codacy license here --->` and `<--- generated_secret --->` placeholders, as these already include instructions in the comments, see below:

https://github.com/codacy/chart/blob/cb7a93cb1474cbdd38e4646deaade1627c3d7472/codacy/values-production.yaml#L11
https://github.com/codacy/chart/blob/cb7a93cb1474cbdd38e4646deaade1627c3d7472/codacy/values-production.yaml#L175-L178

### DIFF
--- a/codacy/values-production.yaml
+++ b/codacy/values-production.yaml
@@ -365,7 +365,7 @@ crow:
       http:
         context: "/monitoring"
     passwordAuth:
-      password: <--- crow-login-password --->
+      password: C0dacy123
 
 codacy-spa:
   replicaCount: 2
--- a/docs/configuration/monitoring.md
+++ b/docs/configuration/monitoring.md
@@ -25,7 +25,7 @@ We highly recommend that you define a custom password for Crow, if you haven't a
     crow:
       config:
         passwordAuth:
-          password: <--- crow password --->
+          password: C0dacy123
     ```
 
 2.  Apply the new configuration by performing a Helm upgrade. To do so execute the command [used to install Codacy](../index.md#helm-upgrade):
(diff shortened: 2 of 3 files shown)


codacy/values-production.yaml
```diff
@@ -365,7 +365,7 @@ crow:
       http:
         context: "/monitoring"
     passwordAuth:
-      password: <--- crow-login-password --->
+      password: C0dacy123
```
pauloribeiro-codacy, line 368: Since the default password is well-known (it is listed in the documentation), it's simpler to make the default password explicit in the .yaml file. Otherwise, customers won't know what to put here until they later read the page on how to set up the monitoring.

PR #532, opened 2020-08-07
fix: Push all rabbitmq prometheusrule in monitoring namespace

joaocosta-codacy:


joaocosta-codacy:
Failing to do this, we will end up with the same prometheus rules across all our namespaces, all of which will trigger the exact same alarms. In the end, we will have a lot of duplicated alarms.
This way, there will always be just one prometheusrule object related to rabbitmq, sitting on the monitoring namespace.

PR #537, opened 2020-08-07
doc: Add Helm 3.3 known incompatibility notice

pauloribeiro-codacy:
Currently, Helm 3.3 is not compatible with MinIO.

Even though we are stating that customers should use Helm 3.2, we're adding a notice to make it explicit that Helm 3.3 will not work.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -45,7 +45,11 @@ Install Codacy on an existing cluster using our Helm chart:
     -   [Helm](https://helm.sh/docs/intro/install/) version 3.2
 
         !!! warning
-            Helm 3.3 is not supported because there is a known incompatibility with the MinIO component used by Codacy.
+            Helm 3.3 is not supported because there is a known incompatibility with the MinIO chart currently used by Codacy.
+            
+            More specifically, **Helm versions compiled with Go 1.14 or later are not supported**. You can check the Go version that was used to compile Helm with `helm version`.
+            
+            This will be fixed in a future version of Codacy Self-hosted.
 
 2.  Create a cluster namespace called `codacy` that will group all resources related to Codacy.
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -44,6 +44,9 @@ Install Codacy on an existing cluster using our Helm chart:
 
     -   [Helm](https://helm.sh/docs/intro/install/) version 3.2
 
+        !!! warning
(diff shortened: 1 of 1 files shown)


pauloribeiro-codacy:
@h314to and @paulopontesm, if this incompatibility is fixed on the next 1.6.0 release, then we shouldn't merge this pull request to `master` and I'll do an hotfix on the `release-1.5.0` branch instead.

dan-codacy:
@h314to Is the incompatibility with just the current chart and it will be likely fixed in future charts? If so, we should change our warning language slightly. Something like "...not compatible with the our current MinIO chart"

pauloribeiro-codacy:
> @h314to Is the incompatibility with just the current chart and it will be likely fixed in future charts? If so, we should change our warning language slightly. Something like "...not compatible with the our current MinIO chart"

I agree, let's tweak the message if we already know this will be fixed soon. :raised_hands:

paulopontesm:
We need this published manually to the live docs, right?

pauloribeiro-codacy:
> We need this published manually to the live docs, right?

Yes, I still need to tweak the message since we already now this will be fixed on the next release, but I'll publish this update on the 1.5.0 and Latest documentation versions. :+1:

docs/index.md
```diff
@@ -44,6 +44,13 @@ Install Codacy on an existing cluster using our Helm chart:
 
     -   [Helm](https://helm.sh/docs/intro/install/) version 3.2
 
+        !!! warning
+            Helm 3.3 is not supported because there is a known incompatibility with the MinIO chart currently used by Codacy.
+            
+            More specifically, **Helm versions compiled with Go 1.14 or later are not supported**. You can check the Go version that was used to compile Helm with `helm version`.
+            
+            This will be fixed in a future version of Codacy Self-hosted.
+
```
pauloribeiro-codacy, line 53: I updated the notice to state that this issue will be fixed in a next version of Codacy Self-hosted (I omitted that it would be fixed on the next version just so we're safe in case something goes wrong).

I also added the [detail about the Go version that causes the incompatibility](https://codacy.slack.com/archives/C8X9SS1H7/p1596830473089300?thread_ts=1596745044.073100&cid=C8X9SS1H7), as it might save a lot of time and trouble to some other customer that is unaware of this.

PR #548, opened 2020-08-13
Update values for components

joaocosta-codacy:


.do/values.yaml
```diff
@@ -34,17 +34,10 @@ global:
...
-    config:
-      dedicatedMax: 20
-      inactivityTimeout: 30
-      analysis:
-        maxFileSizeBytes: 150000
-        # These values are in seconds
-        podDebugInfo: true
-        pluginTimeout:
-          min: 300
-          max: 900
```
joaocosta-codacy, line 47: These are the default values from worker-manager

codacy/values.yaml
```diff
@@ -358,51 +363,10 @@ minio:
 
 worker-manager:
   create: true
-  replicaCount: 1
   fullnameOverride: codacy-worker-manager
   config:
     play:
-      configFile: enterprise.conf
       logFile: logback.xml
     cloudAmqp:
```
paulopontesm, line 370: This could be empty, because of this if: https://bitbucket.org/qamine/worker-manager/src/f1c25d83f602660e095c49d923b082182c1f9a95/.helm/worker-manager/templates/configmap.yaml?at=dev#configmap.yaml-16:17

Maybe put an empty object here and a comment saying that it will use the one from global by default

PR #551, opened 2020-08-18
doc: Mention checking for breaking changes before upgrading Codacy

pauloribeiro-codacy:
Checking for breaking changes should be an explicit step before starting to upgrade Codacy.

### DIFF
--- a/docs/maintenance/upgrade.md
+++ b/docs/maintenance/upgrade.md
@@ -2,7 +2,12 @@
 
 To upgrade Codacy to the latest stable version:
 
-1.  Store all your currently defined configuration values in a file:
+1.  Check the [release notes](/release-notes/) for all Codacy Self-hosted versions between your current version and the most recent version for breaking changes and follow the instructions provided carefully.
+
+    !!! warning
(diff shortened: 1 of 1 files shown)


docs/maintenance/upgrade.md
```diff
@@ -2,7 +2,12 @@
 
 To upgrade Codacy to the latest stable version:
 
-1.  Store all your currently defined configuration values in a file:
+1.  Check the [release notes](/release-notes/) for all Codacy Self-hosted versions between your current version and the most recent version for breaking changes and follow the instructions provided carefully.
```
pauloribeiro-codacy, line 5: If we guarantee that we only add breaking changes when releasing major versions, we can make that explicit here.

docs/maintenance/upgrade.md
```diff
@@ -2,7 +2,12 @@
...
+
+    !!! warning
+        Failing to follow the steps to deal with breaking changes can cause the upgrade to fail or cause problems while Codacy is running.
```
pauloribeiro-codacy, line 8: I'm not sure if we can be more specific here. :thinking:

PR #553, opened 2020-08-21
Add chart version to helm upgrade

dan-codacy:


pauloribeiro-codacy:
I just created #554 so we can continue working on this topic.

PR #559, opened 2020-09-08
docs: add version to upgrade docs

paulopontesm:


docs/maintenance/upgrade.md
```diff
@@ -21,10 +21,17 @@ To upgrade Codacy to the latest stable version:
...
+    ```bash
+    helm repo update
+    ```
+
+5.  Perform the upgrade using the values stored in the file:
 
     ```bash
     helm upgrade codacy codacy-stable/codacy \
                  --namespace codacy \
+                 -- version {{ version }} \
```
pauloribeiro-codacy, line 35: I thought we wouldn't need this flag while upgrading, so that the command upgrades the chart to the latest stable version? Or is it best to specify the current version?

I'm thinking this could be confusing on the following scenario:

- I'm running version 1.5 of the chart, so I open the documentation for this version (https://docs.codacy.com/v1.5/chart/maintenance/upgrade/)
- When I run the command I see on the documentation, I'll be upgrading to version 1.5.0 (`--version 1.5.0`)

What I would expect in this situation would be for the chart to either be upgraded to the latest stable version, or perhaps to the next stable version. :thinking: 

If we do keep the flag, we need to remove the spurious space:

```suggestion
                 --version {{ version }} \
```

docs/maintenance/upgrade.md
```diff
@@ -21,10 +21,17 @@ To upgrade Codacy to the latest stable version:
 
 3.  Review the values stored in the file `codacy.yaml`, making any changes if necessary.
 
-4.  Perform the upgrade using the values stored in the file:
+4.  Get the latest changes from our helm repository
+
+    ```bash
+    helm repo update
```
pauloribeiro-codacy, line 27: To simplify, we could simply include this command as the first line of the code block in the next step (similar to what we have in the initial installation instructions https://docs.codacy.com/chart/#helm-upgrade).

pauloribeiro-codacy (commented):
Please see the comments / questions. :+1:

PR #560, opened 2020-09-08
feature: Remove TLS LetsEncrypt documentation

h314to:


joaocosta-codacy (approved):

pauloribeiro-codacy:
Hi @h314to, could you provide context on the pull request why we are removing this information? It may come in handy in the future if someone asks why it was removed. :+1:

h314to:
Sure @pauloribeiro-codacy.  From what we've seen from self-hosted installations, there isn't a one-size-fits-all solution for adding a certificate and using https. Moreover, this documentation was a bit opinionated regarding the underlying k8s version used and so on, which might lead to some issues when installing. It is safer to to remove this for now, and in the future either provide very basic info about how to upload an existing certificate or just point to upstream documentation.

pauloribeiro-codacy:
The information that we removed from the documentation is still available on our internal Handbook:

https://handbook.dev.codacy.org/engineering/infrastructure/enterprise/tls-ingress.html

PR #573, opened 2020-09-15
Update nginx-ingress to 2.16.0 from kubernetes upstream repository

joaocosta-codacy:


h314to (approved):
I think the documentation also needs to be updated, but maybe that can be done in another PR.

joaocosta-codacy:
Good call. I am updating the documentation as part of this PR. It doesn't involve a lot of effort. 👍

PR #576, opened 2020-09-16
docs: Change release examples to have a 'safer' version number

DReigada:
This is just a small improvement to avoid having people copy paste the examples and release a version 6.0.0 by mistake.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -55,7 +55,7 @@ The Release Manager must create a release candidate branch:
     For example:
 
     ```bash
-    git checkout -b 'release-6.0.0'
+    git checkout -b 'release-x.x.x'
     ```
 
 -   [ ] 4.  Update Dependencies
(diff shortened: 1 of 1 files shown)


pauloribeiro-codacy (approved):
Also consider if we could use a syntax like:

`Major.Minor.Patch`

or, to make it shorter:

`M.m.p`

Although maybe this could be more confusing. :thinking:

joaocosta-codacy (approved):

DReigada:
> Also consider if we could use a syntax like:
> 
> `Major.Minor.Patch`
> 
> or, to make it shorter:
> 
> `M.m.p`
> 
> Although maybe this could be more confusing. 🤔

@pauloribeiro-codacy Although I like the idea, I think it would be more confusing to have it like that

pedrocodacy (approved):
Thank you, I once released a 6.0.0 version by mistake because of this 🥇 ❤️

PR #588, opened 2020-09-18
fix: install latest release on dev

joaocosta-codacy:


joaocosta-codacy:
I am keeping this around since we need to quickly deploy to the dev environment to iterate on tls/https. I will delete this branch once we are done installing things and the environment has stabilised.

PR #590, opened 2020-09-18
update: microk8s version in docs

dan-codacy:
This is related to the 1.16 upgrade by @h314to
He recommends also adding some clarifications about our supported and recommended versions of K8s/microk8s

### DIFF
--- a/docs/infrastructure/microk8s-quickstart.md
+++ b/docs/infrastructure/microk8s-quickstart.md
@@ -27,10 +27,10 @@ Install MicroK8s on the machine:
     sudo apt update && sudo apt install nfs-common -y
     ```
 
-2.  Install MicroK8s from the `1.15/stable` channel:
+2.  Install MicroK8s from the `1.16/stable` channel:
 
     ```bash
-    sudo snap install microk8s --classic --channel=1.15/stable
+    sudo snap install microk8s --classic --channel=1.16/stable
     sudo usermod -a -G microk8s $USER
     sudo su - $USER
     ```
--- a/docs/requirements.md
+++ b/docs/requirements.md
@@ -13,8 +13,8 @@ The cluster running Codacy must satisfy the following requirements:
 
 -   The infrastructure hosting the cluster must be provisioned with the hardware and networking requirements described below
 -   The orchestration platform managing the cluster must be one of:
-    -   [Kubernetes](https://kubernetes.io/) **version 1.14.\*** to **1.18.\***
-    -   [MicroK8s](https://microk8s.io/) **version 1.16**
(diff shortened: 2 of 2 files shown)


dan-codacy:
@pauloribeiro-codacy Should we want to add a microk8s version range to the requirements.md, similar to k8s?
@h314to Do we test against multiple microk8s versions or just 1.15?

pauloribeiro-codacy:
Hi @dan-codacy, and thank you for taking the time to start this update to the documentation.

I will wait for @h314to to advise on what exact MicroK8s versions we should recommend to our customers.

h314to:
I would probably go a step further and recommend a version of k8s, to provide some guidance to users unsure of what version to pick. I'd make it match the one we have tested/used codacy more extensively. Something like:

```markdown
    -   [Kubernetes](https://kubernetes.io/) **version 1.14.\*** to **1.18.\*** (recommended: 1.16)
```

It doesn't mean that it won't work in other versions in that range, but the k8s version is too low it might mean that AWS will drop support for it soon and you'll have to do an upgrade faster that you wanted to. If, on the other hand, the version is too high, we might not have used it so extensively and therefore some unforeseen issues might be lurking around the corner.

pauloribeiro-codacy:
> I would probably go a step further and recommend a version of k8s, to provide some guidance to users unsure of what version to pick. I'd make it match the one we have tested/used codacy more extensively. Something like:
> 
> ```gfm
>     -   [Kubernetes](https://kubernetes.io/) **version 1.14.\*** to **1.18.\*** (recommended: 1.16)
> ```
> 
> It doesn't mean that it won't work in other versions in that range, but if the k8s version is too low it might mean that AWS will drop support for it soon, and you'll have to do an upgrade faster that you wanted to. If, on the other hand, the version is too high, we might not have used it so extensively and therefore some unforeseen issues might be lurking around the corner.
> 
> WDYT?

I think this makes a lot of sense. :+1: 

Are the supported and recommended versions for Kubernetes and MicroK8s the same?

```
    -   [Kubernetes](https://kubernetes.io/) **version 1.14.\*** to **1.18.\*** (1.16 recommended)
    -   [MicroK8s](https://microk8s.io/) **version 1.14.\*** to **1.18.\*** (1.16 recommended)
```

h314to:
LGTM. We should also change the [defaults in the EKS quickstart](https://github.com/codacy/chart/blob/eea1971ee2a28157ae06652133f61b2f37ece2dd/docs/infrastructure/EKS/main/variables.tf#L26). I'll take a look.

h314to (approved):

pauloribeiro-codacy (approved):

PR #599, opened 2020-09-28
fix: Increase resource limits for codacy-tools CY-2568

franciscodua:


franciscodua:
We had to increase the resources in our cloud environment for this service.
So, we wanted to increase the limit requirements for on-premise installations too

pauloribeiro-codacy:
Will this affect the [cluster provisioning requirements](https://docs.codacy.com/chart/requirements/#standard-cluster-provisioning) that we list for customers who are installing Codacy Self-hosted?

franciscodua:
@pauloribeiro-codacy we don't expect the requirements to change since we recently removed the codacy-core service from the chart and didn't reduce the requirements

machadoit (approved):

machadoit:
Even if the requirements increase maybe be bit too high, we keep adding more endpoints and more responsibility for this component. So I think those look fine 👌

joaocosta-codacy (approved):

paulopontesm (approved):

PR #669, opened 2021-03-02
docs: fix pluginTimeout path

machadoit:
This now matches the path at values-production.yaml


jllopes (approved):

lolgab:
Did you notice that the table is not aligned anymore?

pauloribeiro-codacy:
> Did you notice that the table is not aligned anymore?

The table seemed a bit broken some time ago and I fixed it here, maybe this helps? https://github.com/codacy/chart/pull/661

PR #683, opened 2021-04-30
doc: Update supported PostgreSQL version

prcr:
@h314to, is there anything else we need to do before including this newer PostgreSQL version in the Codacy Self-hosted requirements? And should we recommend version 10.15 or 10.14?


h314to:
I think this is fine. There's nothing specific you need to do to run codacy on postgres 10.15

h314to (approved):

h314to:
Some context: We're mostly running 10.14 because that's the version available in aurora, but I don't expect any incompatible changes in 10.15. Moreover, some components using plain Postgres are using 10.15 just fine.

PR #685, opened 2021-05-13
doc: Fix maxFileSizeBytes path

heliocodacy:
maxFileSizeBytes is under workerManager and readme.me was not reflecting that.


prcr:
@joaocosta-codacy, can you confirm that the path in the `values.yaml` file is the correct one?

joaocosta-codacy (approved):

PR #687, opened 2021-05-17
doc: Update Github .pem file instructions

dan-codacy:
Users have been leaving the Begin/End sections in their private key and causing integration failures. I've update the docs, but would like input on the wording to be as clear as possible

### DIFF
--- a/docs/configuration/integrations/github-enterprise.md
+++ b/docs/configuration/integrations/github-enterprise.md
@@ -26,7 +26,7 @@ Follow the instructions below to set up the Codacy Self-hosted integration with 
       app:
         name: "codacy" # GitHub App name
         id: "1234" # GitHub App ID
-        privateKey: "" # Contents of the .pem file without newlines
+        privateKey: "" # Contents of the .pem file without newlines or the Begin/End sections
     ```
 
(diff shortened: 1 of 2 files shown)


docs/configuration/integrations/github-enterprise.md
```diff
@@ -26,7 +26,7 @@ Follow the instructions below to set up the Codacy Self-hosted integration with
       app:
         name: "codacy" # GitHub App name
         id: "1234" # GitHub App ID
-        privateKey: "" # Contents of the .pem file without newlines
+        privateKey: "" # Contents of the .pem file without newlines or the Begin/End sections
```
prcr, line 29: How about:

```suggestion
        privateKey: "" # Contents of the .pem file without newlines or the BEGIN/END lines
```

I wonder if customers have no problem understanding what they have to do regarding the newlines/line breaks part. :thinking:

joaocosta-codacy (approved):

docs/configuration/integrations/github-enterprise.md
dan-codacy: I'm good with this suggestion. It might not be perfect, but this should at least help :-)

PR #691, opened 2021-06-17
fix: Do not send nginx version on http response headers

joaocosta-codacy:


docs/values-files/values-nginx.yaml
```diff
@@ -23,3 +23,6 @@ controller:
     requests:
       cpu: 100m
       memory: 300Mi
+
```
joaocosta-codacy, line 26: This prevents the exposure of the same header on our customers since this file is part of the documentation we provide.

docs/values-files/values-nginx.yaml
```diff
@@ -23,3 +23,7 @@ controller:
...
+  config:
+    # Hides the http response hearders showing the explicit version of nginx
```
mrfyda, line 28:
```suggestion
    # Hides the http response headers showing the explicit version of nginx
```

h314to (approved):

PR #722, opened 2022-02-25
fix: Test with microk8s 1.21 and fix selenium image

h314to:
Removes testing with microk8s 1.14 (which is deprecated). Adds testing with microk8s 1.21. Bumps the chrome web driver to v 3.141.59-20210929, which is also used elsewhere in testing.

### DIFF
--- a/docs/requirements.md
+++ b/docs/requirements.md
@@ -22,8 +22,8 @@ The cluster running Codacy must satisfy the following requirements:
 
 -   The infrastructure hosting the cluster must be provisioned with the hardware and networking requirements described below
 -   The orchestration platform managing the cluster must be one of:
-    -   [Kubernetes](https://kubernetes.io/) **version 1.14.\*** to **1.20.\*** (1.20 recommended)
-    -   [MicroK8s](https://microk8s.io/) **version 1.14.\*** to **1.19.\*** (1.19 recommended)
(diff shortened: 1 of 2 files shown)


medleycodacy (approved):

docs/requirements.md
```diff
@@ -22,8 +22,8 @@ The cluster running Codacy must satisfy the following requirements:
 
 -   The infrastructure hosting the cluster must be provisioned with the hardware and networking requirements described below
 -   The orchestration platform managing the cluster must be one of:
-    -   [Kubernetes](https://kubernetes.io/) **version 1.14.\*** to **1.20.\*** (1.20 recommended)
-    -   [MicroK8s](https://microk8s.io/) **version 1.14.\*** to **1.19.\*** (1.19 recommended)
+    -   [Kubernetes](https://kubernetes.io/) **version 1.15.\*** to **1.20.\*** (1.20 recommended)
```
nfcodacy, line 25: Shouldn't it be "to 1.21"?

lventura-codacy (approved):

docs/requirements.md
h314to: We're just testing in microk8s for now, which is fairly limited. I see it as a first test towards fully supporting 1.21, which will require further testing, and eventually using it ourselves.

nfcodacy (approved):

prcr (approved):
Thanks for also updating the docs! :bookmark_tabs: :+1:

PR #746, opened 2022-07-26
fix: Fix table style in README

lolgab:


prcr (approved):
Nice touch @lolgab! :raised_hands:

PR #763, opened 2022-11-22
bump: Upgrade do k8s to 1.23

tfrgoncalves:


loliveiracodacy (approved):

h314to (approved):

docs/requirements.md
```diff
@@ -22,8 +22,8 @@ The cluster running Codacy must satisfy the following requirements:
 
 -   The infrastructure hosting the cluster must be provisioned with the hardware and networking requirements described below
 -   The orchestration platform managing the cluster must be one of:
-    -   [Kubernetes](https://kubernetes.io/) **version 1.19.\*** to **1.21.\*** (1.21 recommended)
-    -   [MicroK8s](https://microk8s.io/) **version 1.19.\*** to **1.21.\*** (1.19 recommended)
+    -   [Kubernetes](https://kubernetes.io/) **version 1.19.\*** to **1.22.\*** (1.22 recommended)
+    -   [MicroK8s](https://microk8s.io/) **version 1.19.\*** to **1.22.\*** (1.19 recommended)
```
prcr, line 26: Just to confirm, do we still recommend using MicroK8s 1.19?

prcr (approved):
Please check my question, but other than that, looks good! :+1: :shipit:

docs/requirements.md
h314to: I don't think we have a strong technical reason to recommend a more recent version. Microk8s is only used in self-managed instances, so the risk of a provider forcing an upgrade in slim, and I think 1.19 is our most deployed/tested/used version.

PR #767, opened 2022-11-28
fix: Rely on the legacy branch of the release notes

machadoit:


prcr (dismissed):
Thanks for making it work @machadoit! :rocket: :muscle: 

(For future reference, this will make sure that the Makefile target will be using the version of the release notes script available on https://github.com/codacy/release-notes-tools/pull/100.)

lventura-codacy (approved):
Just re-approving. LGTM

PR #769, opened 2022-12-19
fix: use self-hosted components

h314to:


nfcodacy (approved):
I've also checked the two files to ensure that none "legacy" value was forgotten.

tfrgoncalves (approved):

loliveiracodacy (approved):

PR #770, opened 2023-01-11
Fix/docker private access

nfcodacy:


.gitignore
```diff
@@ -25,18 +25,7 @@ chart.iml
...
-*.bak
-*.tmp
-*~
-# Various IDEs
-.project
-.idea/
-*.tmproj
-.vscode/
-**/charts
-chart.iml
```
nfcodacy, line 39: This block was duplicated.

h314to (approved):

PR #780, opened 2023-05-08
bump: Upgrade do cluster to 1.24

tfrgoncalves:
- Bump k8s cluster version from 1.23 to 1.24
- Add terraform plan step to ci
- Add validation to _k8s_cluster_id_ terraform variable, sanity check for tf plan rework
- Bump terraform k8s provider from 1.13.x to 2.20.x
- Bump terraform digital ocean provider from 2.23.x to 2.28.x
- Remove not used k8s provider in db deployment


.do/db-cluster/Makefile
```diff
@@ -6,7 +6,7 @@ TF ?= $(TF_EXTRA_OPTS) terraform
 WORKSPACE_PREFIX ?= postgres
 VARS_FILE ?= deploy.tfvars
 K8S_CLUSTER_REGION ?= fra1
-K8S_CLUSTER_ID ?= $(shell cat ../k8s-cluster/terraform_outputs.json | jq -r '.cluster_id.value')
+K8S_CLUSTER_ID ?= $(shell cat ../k8s-cluster/terraform_outputs.json 2>/dev/null | jq -r '.cluster_id.value' || True && echo 'REPLACE_ME')
```
tfrgoncalves, line 9: Can you think of any improvement to do here?

h314to: how about something like this:
```
doctl kubernetes cluster get codacy-doks-cluster -o json | jq -r '.[].id'
```

.do/db-cluster/variables.tf
```diff
@@ -16,6 +16,10 @@ variable "k8s_cluster_id" {
   description = "The id of the k8s cluster where db connections will come from."
   type = string
   default = "REPLACE_ME"
```
h314to, line 18: You could also refactor this to have as input the cluster name, and then use a [data source](https://registry.terraform.io/providers/digitalocean/digitalocean/latest/docs/data-sources/kubernetes_cluster) to get the cluster, from which you can extract its data, namely, its id

h314to (dismissed):
Looking good, overall. Just left a few comments regarding how to get the cluster id, that may make this more robust.

.do/db-cluster/variables.tf
tfrgoncalves: Thought it would be preferable to keep using the same strategy.

.do/db-cluster/Makefile
tfrgoncalves: Will employ the other approach, thank you.

.do/db-cluster/variables.tf
tfrgoncalves: Implemented.

PR #790, opened 2023-07-06
fix: Increase memory limits components

lolgab:


codacy/values-production.yaml
```diff
@@ -274,17 +274,17 @@ codacy-tools:
       memory: 1Gi
     requests:
       cpu: 0.5
-      memory: 1Gi
```
lolgab, line 277: Codacy tools doesn't require much memory to run

lolgab:
Merging for now. I will tweak them in a follow up PR if there are concerns.

loliveiracodacy:
Didn't see in time to approve, but looks good to me!

PR #827, opened 2026-01-15
infra: upgrade grafana loki

CallMeLeopoldo:
The chart version was causing some issues as well as the config file that we were using for said release. The result of this change is the logs being displayed in Grafana (example was obtained by port-forwarding the grafana service and checking the datasource for loki) 

image

Screenshot 2026-01-15 at 12 06 47

### DIFF
--- a/.do/Makefile
+++ b/.do/Makefile
@@ -119,7 +119,7 @@ deploy_monitoring: #set_cluster_context
 	-kubectl create namespace monitoring
 
 	helm upgrade --install --atomic --timeout 600s loki grafana/loki \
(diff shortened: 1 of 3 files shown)


jorgebraz (commented):
Does this change impact existing SH customers in any way? 
For example: will customers have to change anything in their installations so logging works for them?

CallMeLeopoldo:
> Does this change impact existing SH customers in any way? For example: will customers have to change anything in their installations so logging works for them?

Nope, this change only affects customers that are doing a fresh (and first) install of the SH instance.

jorgebraz (approved):
LGTM 🚢

